PR #8, opened 2022-11-08
Rename DIFFERENCE.md to difference.md

tabcat:
fixes broken link in README.md and matches style of other document filenames.


marcus-pousette:
great!

Issue #14, opened 2022-11-14
Query protocol bloaty


marcus-pousette:
- AnySearch module can be merge into DQuery and Document and DString and be removed
- Search results from any DQuery should return context (the address)
- Search should allow for querying any state "created at" and "last edited at"

marcus-pousette:
Blocked by #11

PR #17, opened 2022-11-15
Remove AnySearch module

marcus-pousette:
- Simplified query interface where context related things are moved to ```DQuery``` and rest of the query spec. is absorded in each specific data structure (```Documents``` and ```DString```)


marcus-pousette:
Partially solved: #14

Issue #20, opened 2022-11-19
Multiple signers per commit


marcus-pousette:
Good solutions to #19 #15 expect that a commit can be signed by multiple parties. 

For #19 we want a trusted "clock service" to sign root commits that they have correct timestamps. 

For #15 we can keep sign messages with 2 identities, 1 identity that allows the message to be stored on a replicator. 1 identity that proves that for the end receiver that you are you

PR #21, opened 2022-11-20
Multiple signers

marcus-pousette:
- Allow multiple signers for commits
- ClockService that can sign entries to verify that the timestamp is accurate
- Rename Query to RPC


marcus-pousette:
Solves #19 #20

PR #44, opened 2023-01-26
Querying documents for null/undefined fields

threshold862543:
This adds a document query "FieldIsEmptyQuery" that allows you to specify a field to query if the value of that field is (or isn't) null or undefined:

`new FieldIsEmptyQuery({key: 'favoritefood', isempty: true})`

Please let me know if the name (or anything else) should be changed to fit the current naming scheme, or if this should be gone about in a different way.

### DIFF
--- a/packages/programs/data/document/src/document-index.ts
+++ b/packages/programs/data/document/src/document-index.ts
@@ -311,7 +311,7 @@ export class DocumentIndex<T> extends ComposableProgram {
                               if (f instanceof FieldIsEmptyQuery) {
                                   const value: bigint | number | undefined = fv;
 
-                                  return value === null || value === undefined;
+                                  return !(f.isempty && (value === null || value === undefined))
                               }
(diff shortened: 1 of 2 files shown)


marcus-pousette:
Thank you for the PR :) 

- I think the name should be FieldMissingQuery, because you want to capture both undefined and null. The "Is" does not bring that not much more value
- You would have to add tests for me to be able to merge it (test that make sure both undefined and null works as expected, but it does not capture "" as missing). 

However I am working on some refactor things at the moment might be worth postponing work (adding tests) on this PR until that is done

marcus-pousette:
I closed this as resolved as I included the update in the latest perf. work. [Example missing query](https://github.com/dao-xyz/peerbit/blob/c50960f376675e42549a8a28c4f69b6594c9d021/packages/programs/data/document/src/__tests__/index.integration.test.ts#L414)

Issue #213, opened 2023-10-04
Support for symmetric key within EncryptedThing and DecryptedThing


benjaminpreiss:
It would be great to have support for both asymmetric and symmetric crypto in EncryptedThing and DecryptedThing.

marcus-pousette:
Done ! :)  See comment https://github.com/dao-xyz/peerbit/pull/215#issuecomment-1925295426

PR #214, opened 2023-10-04
Add symmetric capabilities to DecryptedThing

benjaminpreiss:
#213

### DIFF
--- a/packages/utils/crypto/src/encryption.ts
+++ b/packages/utils/crypto/src/encryption.ts
@@ -27,7 +27,10 @@ export abstract class MaybeEncrypted<T> {
 	}
 
 	decrypt(
-		keyOrKeychain?: Keychain | X25519Keypair
+		keyOrKeychain?:
+			| Keychain
+			| X25519Keypair
+			| Uint8Array /* Comment: The last is added */
 	): Promise<DecryptedThing<T>> | DecryptedThing<T> {
 		throw new Error("Not implemented");
 	}
@@ -43,6 +46,24 @@ export abstract class MaybeEncrypted<T> {
 	}
 
 	abstract get byteLength(): number;
+}
+
+type EncryptAsymmetricParameters = {
+	x25519Keypair: X25519Keypair;
+	receiverPublicKeys: (X25519PublicKey | Ed25519PublicKey)[];
+};
+
+type EncryptSymmetricParameters = { symmetricKey: Uint8Array };
+
+type EncryptReturnValue<Parameters, T> =
+	Parameters extends EncryptSymmetricParameters
+		? EncryptedSymmetricThing<T>
+		: EncryptedThing<T>;
+
+function isEncryptSymmetricParameters(
+	parameters: EncryptSymmetricParameters | EncryptAsymmetricParameters
+): parameters is EncryptSymmetricParameters {
+	return (parameters as EncryptSymmetricParameters).symmetricKey !== undefined;
 }
 
 @variant(0)
@@ -69,50 +90,67 @@ export class DecryptedThing<T> extends MaybeEncrypted<T> {
 		return deserialize(this._data, clazz);
 	}
 
-	async encrypt(
-		x25519Keypair: X25519Keypair,
-		...receiverPublicKeys: (X25519PublicKey | Ed25519PublicKey)[]
-	): Promise<EncryptedThing<T>> {
+	async encrypt<
+		Parameters extends EncryptAsymmetricParameters | EncryptSymmetricParameters
+	>(
+		parameters: EncryptAsymmetricParameters | EncryptSymmetricParameters
+		/* Comment: instead of 
+        x25519Keypair: X25519Keypair,
+		...receiverPublicKeys: (X25519PublicKey | Ed25519PublicKey)[] */
+	): Promise<
+		EncryptReturnValue<Parameters, T>
+	> /* Comment: instead of Promise<EncryptedThing<T>> */ {
 		const bytes = serialize(this);
-		const epheremalKey = sodium.crypto_secretbox_keygen();
+		const epheremalKey = isEncryptSymmetricParameters(parameters)
+			? parameters.symmetricKey
+			: sodium.crypto_secretbox_keygen();
+		/* Comment: instead of const epheremalKey = sodium.crypto_secretbox_keygen(); */
 		const nonce = randomBytes(NONCE_LENGTH); // crypto random is faster than sodim random
 		const cipher = sodium.crypto_secretbox_easy(bytes, nonce, epheremalKey);
 
-		const receiverX25519PublicKeys = await Promise.all(
-			receiverPublicKeys.map((key) => {
-				if (key instanceof Ed25519PublicKey) {
-					return X25519PublicKey.from(key);
-				}
-				return key;
-			})
-		);
+		if (!isEncryptSymmetricParameters(parameters)) {
+			const { receiverPublicKeys, x25519Keypair } = parameters;
+			const receiverX25519PublicKeys = await Promise.all(
+				receiverPublicKeys.map((key) => {
+					if (key instanceof Ed25519PublicKey) {
+						return X25519PublicKey.from(key);
+					}
+					return key;
+				})
+			);
 
-		const ks = receiverX25519PublicKeys.map((receiverPublicKey) => {
-			const kNonce = randomBytes(NONCE_LENGTH); // crypto random is faster than sodium random
-			return new K({
-				encryptedKey: new CipherWithNonce({
-					cipher: sodium.crypto_box_easy(
-						epheremalKey,
-						kNonce,
-						receiverPublicKey.publicKey,
-						x25519Keypair.secretKey.secretKey
-					),
-					nonce: kNonce
-				}),
-				receiverPublicKey
+			const ks = receiverX25519PublicKeys.map((receiverPublicKey) => {
+				const kNonce = randomBytes(NONCE_LENGTH); // crypto random is faster than sodium random
+				return new K({
+					encryptedKey: new CipherWithNonce({
+						cipher: sodium.crypto_box_easy(
+							epheremalKey,
+							kNonce,
+							receiverPublicKey.publicKey,
+							x25519Keypair.secretKey.secretKey
+						),
+						nonce: kNonce
+					}),
+					receiverPublicKey
+				});
 			});
-		});
-
-		const enc = new EncryptedThing<T>({
+			const enc = new EncryptedThing<T>({
+				encrypted: new Uint8Array(cipher),
+				nonce,
+				envelope: new Envelope({
+					senderPublicKey: x25519Keypair.publicKey,
+					ks
+				})
+			});
+			enc._decrypted = this;
+			return enc as EncryptReturnValue<Parameters, T>;
+		}
+		const enc = new EncryptedSymmetricThing<T>({
 			encrypted: new Uint8Array(cipher),
-			nonce,
-			envelope: new Envelope({
-				senderPublicKey: x25519Keypair.publicKey,
-				ks
-			})
+			nonce
 		});
 		enc._decrypted = this;
-		return enc;
+		return enc as EncryptReturnValue<Parameters, T>;
 	}
 
 	get decrypted(): DecryptedThing<T> {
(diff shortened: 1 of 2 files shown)


packages/utils/crypto/src/encryption.ts
```diff
@@ -374,3 +412,138 @@ export class EncryptedThing<T> extends MaybeEncrypted<T> {
...
+		if (!this._decrypted) {
+			throw new Error(
+				"Entry has not been decrypted, invoke decrypt method before"
+			);
+		}
+		return this._decrypted;
+	}
+
+	async decrypt(
+		keyResolver?: Uint8Array /* Comment: instead of Keychain | X25519Keypair */
```
marcus-pousette, line 443: I wonder whether the Keychain should be improved to also be able to export Ephemeral keys based on their hashes or something

But this is about solving the problem 

- How do I know I have decrypted something correctly?
- How do I know that I can decrypt it before decrypting it? (Lets say I have 1000 ephemeral keys in the keystore, I can not test all of them)

benjaminpreiss: Interesting thought... So that means one would also include an Envelope within `EncryptedSymmetricThing` that includes the hash of the SymmetricKey?

benjaminpreiss: ... Well maybe the naming would not be envelope any more as there is no signature.

benjaminpreiss: Regarding your questions:

> How do I know I have decrypted something correctly?

I think certainty can be provided by successful deserialization?

> How do I know that I can decrypt it before decrypting it? (Lets say I have 1000 ephemeral keys in the keystore, I can not test all of them)

When we include the hash of the ephemeral key used for encryption in the "envelope", we have a linear lookup time for the question "Can I decrypt?".

marcus-pousette: > I think certainty can be provided by successful deserialization?

There might be scenarios where the decrypted ciphertext still can deserialize succesfully into your object even though the decryption is invalid. Lets say it is not something you can trust on 100%

But this page is pretty good

https://libsodium.gitbook.io/doc/secret-key_cryptography/secretbox


in js you can use it 
`libsodium.crypto_secretbox_easy`
(libsodium is already a dependency of peerbit crypto). 

Does the integrity check. I think it bundles some authentication data in the cipher text that depends on the nonce and the secret key used. So that issue would be solved (but one has to test) that this will be sufficient for the verifying it has been decrypted correctly.


> When we include the hash of the ephemeral key used for encryption in the "envelope", we have a linear lookup time for the question "Can I decrypt?".

That would be an easy way to go about it. But it is not standard thing to do it seems.. (I am lookig around on the web and ask the chatbot). But it feels very intuitive

An alternative thing would be that this is on the app level. So you would a have a keychain locally which is a KV store, and you would put your symmetric keys in there with well chosen IDs, like it could be the db address of the store you want to encrypt. So basically when you want to decrypt a db you would go into this keychain and look for this key that has an id which is the db address and then if it exist, use it to decrypt entries. 

Another alternative would be that you would have a shared keychain which is a document store for ephemeral keys which is encrypted with publickey encryption (the encryption you get default with the document store now). With that you can have each document to have an id which you somehow connect with the id you use for the messages you encrypt

benjaminpreiss: > That would be an easy way to go about it. But it is not standard thing to do it seems.. (I am lookig around on the web and ask the chatbot). But it feels very intuitive

Yes, that might be. I don't think it is insecure though to do so (if key and salt are sufficiently large).

> An alternative thing would be that this is on the app level. So you would a have a keychain locally which is a KV store, and you would put your symmetric keys in there with well chosen IDs, like it could be the db address of the store you want to encrypt. So basically when you want to decrypt a db you would go into this keychain and look for this key that has an id which is the db address and then if it exist, use it to decrypt entries.

> Another alternative would be that you would have a shared keychain which is a document store for ephemeral keys which is encrypted with publickey encryption (the encryption you get default with the document store now). With that you can have each document to have an id which you somehow connect with the id you use for the messages you encrypt in another db (this solution would also help you send your ephemeral keys to the other party)

These would be solutions that could be leveraged additionally to ensure successful decryption?

marcus-pousette: > Yes, that might be. I don't think it is insecure though to do so (if key and salt are sufficiently large). Would you still name the bundle of salt + key hash Envelope ?

This issue if that if the hash of the symmetric key depends on the salt then the lookup of the key in your K/V keychain can not be done(?). I assume the keychain stores keys with their hashes as key and value as the secret. If I get a the hash of a key but it is salted I am not sure what I am going to do with it. 

Also you could just used the nonce as the salt if you want to simplify things (given salt would still be necessary). I wonder whether you would have the same outcome of "salt" as with just generate a new symmetric key for each message? In that case if one gets lost, it can not nevertheless be used to decrypt other messages

marcus-pousette: > These would be solutions that could be leveraged additionally to ensure successful decryption?

Yeap. Perhpas this is more about key exchange also. But the reason why I bring this up is that EncryptedThing does the keyexchange in the message and does symmetric encryption with the key that is exchanged.

And EncryptedSymmetricThing is basically the same thing as EncryptedThing except that there is no key exchange using PublicKeyEncrypted. It is basically a "Detached" version of EncryptedThing

marcus-pousette: Actually (!) the simplest way around this problem would just be to rename the 

"Envelope" that exist in the code now to `PublicKeyEnvelope` (variant 0)

and then do a new one like 

`HashedKeyEnvelope` (variant 1)  which stores the hash of the symmetric key

Then both solutions would still use the EncryptedThing class but the envlope would be different

benjaminpreiss: > > Yes, that might be. I don't think it is insecure though to do so (if key and salt are sufficiently large). Would you still name the bundle of salt + key hash Envelope ?
> 
> This issue if that if the hash of the symmetric key depends on the salt then the lookup of the key in your K/V keychain can not be done(?). I assume the keychain stores secrets with their hashes as key and value as the secret. If I get a the hash of a key but it is salted I am not sure what I am going to do with it.
> 
> Also you could just use the nonce as the salt if you want to simplify things (given salt would still be necessary). I wonder whether you would have the same outcome of "salt" as with just generate a new symmetric key for each message? In that case if one gets lost, it can not nevertheless be used to decrypt other messages

Ah, I would actually store the hashed value of key + salt in the keystore...

marcus-pousette: pseudo code

```js

abstract class AbstractEnvelop {}

@variant(0)
class PublicKeyEnvelope { ... Envlope code for public key key exchange .. }

@variant(1)
class HashedKeyEnvelope {  
@field({type: fixed('u8',32}) 
 hash: Uint8array
 }

```

marcus-pousette: > Ah, I would actually store the hashed value of key + salt in the keystore...

Since the salt (or nonce) is determined by the sender and you you should already have the key in your keystore. This will not work. Lets say we decide on a  symmetric key in person and we both type that key into the keystore. Then later when we do messaging we can not make the hash of the symmetric key no other than the hash of the secret key if we want a simple lookup in the keystore.

benjaminpreiss: > Since the salt (or nonce) is determined by the sender at the moment of writing the message and you should already have the key in your keystore: This will not work. Lets say we decide on a symmetric key in person and we both type that key into the keystore. Then later when we do messaging we can not make the hash of the symmetric key no other than the hash of the secret key if we want a simple lookup in the keystore.

Not sure if I agree honestly - I understand symmetric key encryption as a mechanism where the secret has to be distributed to the recipient via another channel (As both sender and recipient need the same key to encrypt / decrypt). It doesn't matter then whether the recipient has the key at the time of encryption on the sender side, as long as he has the key when he wants to decrypt.

The salt / nonce is then stored unencrypted alongside the key hash for the recipient to retrieve

benjaminpreiss: > pseudo code
> 
> ```js
> abstract class AbstractEnvelope {}
> 
> @variant(0)
> class PublicKeyEnvelope extends AbstractEnvelope { ... Envlope code for public key key exchange .. }
> 
> @variant(1)
> class HashedKeyEnvelope extends AbstractEnvelope {  
> @field({type: fixed('u8',32}) 
>  hash: Uint8array
>  }
> ```

Beautiful idea!

marcus-pousette: > I understand symmetric key encryption as a mechanism where the secret has to be distributed to the recipient via another channel (As both sender and recipient need the same key to encrypt / decrypt).

Yes

>  It doesn't matter then whether the recipient has the key at the time of encryption on the sender side, as long as he has the key when he wants to decrypt.

Yes we are on the same side of this. 

I am perhaps just talking about solving the question

- Can I decrypt this message? 

Imagine a solution without where the hash of the secret key is not salted then the flow would be 

Received message -> Retrieve secret key hash in message envelope -> Lookup in my keystore which is a K/V or Map where the key is the hash of the secret. i.e. Keystore.get(secret key hash)  O(1)  -> not undefined ? then we can open the message ! 


Now if the hash of the secret key is salted then 

Received message -> Retrieve SALTED secret key hash in message envelope -> ❌ Lookup in my keystore which is a K/V or Map where the key is the hash of the secret. i.e. Keystore.get(key hash)  O(1) ❌ 

The third step fails because the lookup table in my keystore expects the keys to be the hash of the secret key, not the hash of the secretkey + salt

benjaminpreiss: > Received message -> Retrieve SALTED secret key hash in message envelope -> ❌ Lookup in my keystore which is a K/V or Map where the key is the hash of the secret. i.e. Keystore.get(key hash) O(1) ❌

Well, I think I would exchange the symmetric key always with the salt it is used with, so that we actually have multiple hashes. My scenario would be:

1. Sender generates salt
2. Sender adds salt to symm key and hashes the result
3. Sender sends the symm key and the previously generated hash on side channel
4. Recipient stores hash -> symm key mapping in KV store

The recipient could receive the same symm key multiple times but for different salts resulting in multiple entries in KV store pointing to the same symm key

But I understand your concern. For this to happen, the recipient needs to receive the hash value alongside the symm key to answer the question "Can I decrypt?".

When the recipient for some unknown reason only knows the symm key, he cannot sufficiently answer that question in this scenario.

The question remains - what can we do?

marcus-pousette: > When the recipient for some unknown reason only knows the symm key, he cannot sufficiently answer that question in this scenario.

So what I want to continue to say on this topic is that if you want to receiver to have all necessary data to lookup the secretkey from the keystore you practically need to talk on the sidechannel for every message (?) (given that the salt is unique for every message). If you are to talk about the encryption for about every message you might just want to generate a new symmetric key for every message altogether and use the sidechannel to send that over instead. 

Using salts is generally done with a slow hashfunction for passwords which have a low entropy. This prevents someone from bruteforcing passwords given that they know they salt and the password + salt hash. But in this case the "password" is high entropy and there is no possibility that an actor will try to generate all 32 byte symmetric keys to match the hash. The only thing would be that if the symmetric key gets lost, then a middle man could have a easy way of figuring out what messages the middle man can decrypt. Having a salt here would not help anything, or making anything harder if the secret key gets lost. For the same reason as the salt for a password db does not help anything if the password in plain text gets lost


> The question remains - what can we do?

i would just keep it simple! 

 > HashedKeyEnvelope 
 just make it
 
  { hash: Uint8Array } 
  
  
 Pros: 
-  Will solve a problem where you want to have encryption but share keys on a sidechannel instead of inside the message itself (i.e. use the PublicKeyEnvelope) 
 
- The hash will help a receiver to figure out if they have a key in the keystore
 
- Noone can find out what the symmetric key is given the hash

- I can loop over all messages efficiently and figure out what messages I can decrypt, if not I can use a sidechannel to ask for the symmetric key by refering to its hash

Cons:
- Someone from the outside can fiigure out that all messages are encrypted with the same key

benjaminpreiss: Agreed! That's how I will implement it.

benjaminpreiss: Would you then also just keep one `EncryptedThing` or split up in `EncryptedThing` and `SymmetricEncryptedThing`?

What type would you suggest for the unified `EncryptedThing` `encrypt` function?
```ts
async decrypt(
	keyResolver?: Keychain | X25519Keypair
): Promise<DecryptedThing<T>>
```

packages/utils/crypto/src/encryption.ts
```diff
@@ -240,13 +313,13 @@ export class EncryptedThing<T> extends MaybeEncrypted<T> {
 	@field({ type: Uint8Array })
 	_nonce: Uint8Array;
 
-	@field({ type: Envelope })
-	_envelope: Envelope;
+	@field({ type: PublicKeyEnvelope })
+	_envelope: PublicKeyEnvelope;
```
marcus-pousette, line 317:
```suggestion
	@field({type: AbstractEnvelope})
        _envelope: PublicKeyEnvelope | HashedKeyEnvelope
```

packages/utils/crypto/src/encryption.ts
```diff
@@ -240,13 +313,13 @@ export class EncryptedThing<T> extends MaybeEncrypted<T> {
...
 
 	constructor(props?: {
 		encrypted: Uint8Array;
 		nonce: Uint8Array;
-		envelope: Envelope;
+		envelope: PublicKeyEnvelope;
```
marcus-pousette, line 322:
```suggestion
		envelope: PublicKeyEnvelope | HashedKeyEnvelope;
```

packages/utils/crypto/src/encryption.ts
marcus-pousette: > Would you then also just keep one `EncryptedThing` or split up in `EncryptedThing` and `SymmetricEncryptedThing`?
> 
> What type would you suggest for the unified `EncryptedThing` `encrypt` function?
> 
> ```ts
> async decrypt(
> 	keyResolver?: Keychain | X25519Keypair // | Uint8Array like this?
> ): Promise<DecryptedThing<T>>
> ```

One encrypted thing, because later lets say the Keystore can both store Symmetric keys and X25519 Keypairs then you can do 


```js
await encryptedThing.decrypt(myKeystore) 
```

without knowing the details of what the Envelope is.

But there is a problem now with the Keystore class is that it depends on the libp2p Keystore object which only can store PeerIds.. and it will also be dropped soon 

https://github.com/libp2p/js-libp2p/pull/2084

marcus-pousette: There might be a sidequest to this to rework the keystore to have following features
- Can store all keytypes necessary
- Can look them up by their IDs or by their PublicKeys or their secret key hashes

marcus-pousette:
See https://github.com/dao-xyz/peerbit/pull/215#issuecomment-1925295426

Closing

PR #217, opened 2023-10-09
216 build keychainprogram

benjaminpreiss:


packages/programs/program/src/keychain.ts
```diff
@@ -0,0 +1,219 @@
...
+		// If owner is me, create a new permanent key with
+		if (keys.length === 0) {
+			//return compare(this.identity.publicKey.bytes, owner) ?
+		}
+		return keys[0].keypair.decrypt(this.decryptProvider);
+	}
+
+	// TODO: Unsure about encryptProvider here honestly - should this be layer 1 keychain?
+	async open(parameters: {
+		identity: Identity;
```
benjaminpreiss, line 146: identity already exist if you do

`this.identity` inside open (it is a global thing on 'this')

packages/programs/program/src/keychain.ts
```diff
@@ -0,0 +1,219 @@
...
+			| EncryptedSecp256k1Keypair;
+		revoked?: boolean;
+	}) {
+		this.keypair = parameters.keypair;
+		this.owner = parameters.owner;
+		this.revoked = parameters.revoked ?? false;
+	}
+}
+
+export class KeychainProgram extends Program {
```
marcus-pousette, line 71: `implements Keychain` maybe?  from '@peerbit/keychain'

packages/programs/program/src/keychain.ts
```diff
@@ -0,0 +1,219 @@
...
+	@field({ type: Documents<EncryptedExtendedKey> })
+	keys: Documents<EncryptedExtendedKey>;
```
marcus-pousette, line 73: When PR https://github.com/dao-xyz/peerbit/pull/215 is merged it should in practice just be possible to just do 

```js
@field({ type: Documents<EncryptedExtendedKey> })
keys: Documents<Ed25519Keypair>; // or something else secret
```


and later just do 

```
sharedKeychain.keys.put(keypair, { encryption : { hash: ` the hash of the secret key you want to use` } })
```

marcus-pousette: or, well it depends if you want to have a a public and private part of the document of course. Then you need the `EncryptedExtendedKey` type

benjaminpreiss: Hmmm, but isn't the problem that the `encryption` property on the `put` method encrypts the whole key? How would I then find the public key of someone else that I want to encrypt something for, if the public key is encrypted as well?

benjaminpreiss: Also, how can I mark keys as revoked?

benjaminpreiss: And I have also been thinking - sometimes, a user I want to encrypt something for has not generated a keypair yet. So at that time of speaking, I want to generate a keypair for him that is only single use.

marcus-pousette: > property on the put method encrypts the whole key? How would I then find the public key of someone else that I want to encrypt something for, if the public key is encrypted as well?

Ah yee for this to work the decryption has to be performed by keys that exist in a lower level keychain. When you use the put method there PublicKeyEnvelope (in the crypto lib) will leak the publickeys of the parties that can decrypt the message, which allows you to figure out if you can decrypt a message.


I think it might be easier to reason about stuff if you would separate "encryption" utilties from this shared-keychain. So this keychain stores keys that are encrypted, but the keys for providing the encryption  for storing these keys for this keychain perhaps needs to originate from a lower level keychain. 

It could in practice also be possible that keys in this keychain are allso encrypted by keys that also exist in this keychain, but then you need to make sure you are loading the database in the right order kind of (which is possible in practice there is a oplog behind the scenes that allows you to link commits)

marcus-pousette: If you want to revoke keys I think you need a revoked flag on the keypair object, or something like that. Else you could in theory also just use the delete function of the document store and assumes keys are only valid if they exist in the keychain

PR #291, opened 2024-07-15
feat!: Persistant indexer

marcus-pousette:
- Generic indexing interface
     - SQLite impl
     - Hashmap impl
- Use indexing framework for log entry management, replication ranges and document store index
- Migrate from jest to mocha


threshold862543:
Nice!!

Issue #325, opened 2024-09-06
Possibility to limit bandwidth up/down


marcus-pousette:
In a few applications developers might want to limit upload and download speed, e.g. to make sure some bandwidth is reserved for other apps

saul-jb:
It would be great to be able to do this on a per-item level, e.g. the files demo - prioritize syncing of one large file over another.

PR #544, opened 2026-01-17
feat(shared-log): add sync prioritization

### LINKED ISSUE #521: Item Prioritisation
Probably related to #325, is there anyway in the current version to have prioritisation of items to be synced? I.e. in the file demo your peer might be configured to replicate 50% but you might want certain items downloaded sooner than others - maybe based of a numerical priority metadata attached to the file? I could see that maybe you could get something like this working with iteration syncing and sort but it would be nice to be done in a way that works with the replication ranges and limit configuration.

marcus-pousette:
Adds an opt-in sync prioritization mechanism for shared-log syncing.

### Motivation
When a peer joins and is missing a lot of entries, the default sync order can delay *time-to-first-useful-data*.
Typical examples are apps that mix small “index/manifest” entries with lots of bulk/history entries (e.g. file share: manifest + directory listing first, then file chunks).

### What’s included
- `sync.priority(entry): number` — ranks entries (higher first) when the simple synchronizer sends `RequestMaybeSync`.
- `sync.maxSimpleEntries` — when using rateless IBLT for large batches, pre-sync up to **N** highest-priority entries (plus range-boundary entries) via the simple synchronizer before the IBLT exchange.
- `Documents.open(...)` forwards `sync` (and other shared-log options) through to the underlying shared-log.

### Usage example (Documents)
```ts
import { Documents } from "@peerbit/document";

const hot = new Set<string>();
// e.g. after a query / UI selection, add the hashes you want to arrive ASAP:
// hot.add(resultHash)

const docs = await peer.open(new Documents<MyDoc>({ id }), {
  args: { type: MyDoc },

  sync: {
    maxSimpleEntries: 32,
    priority: (entry) => (hot.has(entry.hash) ? 100 : 0),
  },
});
```

Notes:
- `priority` should be fast and side-effect free.
- For many apps (like file-share), `priority` is best derived from your replication domain/coordinates so all peers make the same choice.

Fixes #521
Refs #325

### DIFF
--- a/packages/programs/data/document/document/src/program.ts
+++ b/packages/programs/data/document/document/src/program.ts
@@ -305,10 +305,18 @@ export class Documents<
 			trim: options?.log?.trim,
 			replicate: options?.replicate,
 			replicas: options?.replicas,
+			respondToIHaveTimeout: options?.respondToIHaveTimeout,
+			sync: options?.sync,
+			syncronizer: options?.syncronizer,
+			timeUntilRoleMaturity: options?.timeUntilRoleMaturity,
+			waitForReplicatorTimeout: options?.waitForReplicatorTimeout,
+			waitForPruneDelay: options?.waitForPruneDelay,
+			distributionDebounceTime: options?.distributionDebounceTime,
 			domain: (options?.domain
 				? (log: any) => options.domain!(this)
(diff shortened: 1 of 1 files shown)


marcus-pousette:
Follow-up tracking issue: #545

Issue #548, opened 2026-01-18
shared-log: rateless-iblt rebuild cost per sync session


marcus-pousette:
In the current rateless-IBLT sync implementation we appear to rebuild the IBLT “sketch” (encoder/decoder) for every sync session, which may be expensive when many peers join or when we do repeated partial syncs.

### What happens today
- **Sender side:** `RatelessIBLTSynchronizer.onMaybeMissingEntries(...)` creates a new `EncoderWrapper` and calls `add_symbol(...)` for every coordinate in the batch.
- **Receiver side:** handling `StartSync` calls `buildEncoderOrDecoderFromRange(...)`, which does an `entryIndex.iterate(...).all()` over the selected range and builds a fresh `DecoderWrapper` by adding every local symbol in that range.

Refs:
- `packages/programs/data/shared-log/src/sync/rateless-iblt.ts`

### Why this might be a problem
- When a **new peer subscribes**, many existing peers can end up syncing lots of entries to it. Each sync session triggers a full rebuild of encoder/decoder state (CPU + index IO).
- Even with `minSyncIbltSize` (currently hard-coded `333`) to avoid IBLT for small batches, the rebuild cost for large batches could dominate join time.

There’s already a related concern noted in the code:
- `packages/programs/data/shared-log/src/index.ts` has a TODO comment about syncing overhead for new (immature) replicators.

### Possible improvements
1. **Benchmark / measure**
   - Add benchmarks for encoder/decoder build time at ~1k/10k/100k entries and a “N peers join” scenario.
2. **Cache local symbol lists per range**
   - Cache the list of local `hashNumber`s for `(start,end)` ranges (TTL + invalidation on entry add/remove) so `buildEncoderOrDecoderFromRange` can avoid re-scanning the index.
3. **Concurrency limits / backpressure**
   - If many `StartSync` messages arrive concurrently, cap parallel decoder builds.
4. **Make thresholds configurable**
   - Expose `minSyncIbltSize` / related knobs so CPU-limited devices can prefer the simple synchronizer.

### Questions
- Is the intent that IBLT rebuild is "cheap enough" in practice due to typical segment sizes, or do we expect this to be a bottleneck under churn?
- Should we align IBLT ranges with stable replication segments so caching becomes straightforward?

marcus-pousette:
Bench groundwork in PR #549: adds packages/programs/data/shared-log/benchmark/rateless-iblt-rebuild.ts.

Run:

```bash
[...]

PR #550, opened 2026-01-18
perf(shared-log): cache local IBLT range encoders

marcus-pousette:
Caches the local (receiver-side) symbol set for rateless-IBLT `StartSync` ranges.

### Why
Building the local decoder currently rebuilds the symbol set by scanning the index and calling `add_symbol(...)` for every sync session. Under churn / repeated syncs on the same range this can be expensive.

### What changed
- Adds a small, bounded cache of `EncoderWrapper`s keyed by `(start1,end1,start2,end2)` and a local "version".
- `StartSync` now uses `clone().to_decoder()` from the cached encoder when possible, avoiding a full rebuild.
- Cache is invalidated on `onEntryAdded` / `onEntryRemoved` and freed on `close()`.
- Adds unit tests verifying cache reuse + invalidation.

### How to verify
- Unit test:
  ```bash
  node ./node_modules/aegir/src/index.js run test \
    --roots ./packages/programs/data/shared-log -- -t node \
    --grep "rateless-iblt-syncronizer cache"
  ```
- Benchmark (cold vs warm decoder build):
  ```bash
  cd packages/programs/data/shared-log
  RIBLT_SIZES="1000,10000" RIBLT_WARMUP=5 RIBLT_ITERATIONS=20 \
    node --loader ts-node/esm ./benchmark/rateless-iblt-startsync-cache.ts
  ```
  Expect the "warm" case to be significantly faster than "cold".

Refs #548

### DIFF
--- a/packages/programs/data/shared-log/benchmark/rateless-iblt-startsync-cache.ts
+++ b/packages/programs/data/shared-log/benchmark/rateless-iblt-startsync-cache.ts
@@ -1,0 +1,118 @@
+import { Cache } from "@peerbit/cache";
+import { ready as ribltReady } from "@peerbit/riblt";
+import { Bench } from "tinybench";
+import { createNumbers, type Numbers } from "../src/integers.js";
+import { RatelessIBLTSynchronizer } from "../src/sync/rateless-iblt.js";
+
+// Run with:
+//   cd packages/programs/data/shared-log
+//   RIBLT_SIZES="1000,10000,50000" RIBLT_WARMUP=5 RIBLT_ITERATIONS=20 \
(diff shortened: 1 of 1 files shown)


marcus-pousette:
I ran the new benchmark locally (very low iterations) and it shows the expected behavior: warm cache hits are much faster than cold rebuilds.

Command:
```bash
cd packages/programs/data/shared-log
RIBLT_SIZES=1000 RIBLT_WARMUP=1 RIBLT_ITERATIONS=3 \
  node --loader ts-node/esm ./benchmark/rateless-iblt-startsync-cache.ts
```

On my run `StartSync local decoder (warm)` was ~6x faster than `cold` at n=1000 (and `after invalidation` matches `cold`).

PR #593, opened 2026-02-06
Fix/shared log handshake flakes

Faolain:
## Why

Pubsub was flaky and has race conditions with topic subscribing. #589 solved that but that surfaced another already existing race condition within the shared-log itself which should be handled in this PR. Read that PR for the full context of the problem (in the dApp it would surface as inability to make connections with a peer, "flaky connection") Also please skip to the bottom for an additional more targeted approach that was tried with PubSub https://github.com/Faolain/peerbit/pull/7 which I'm thinking to stack with this PR if you agree.

## Context

PR #589 (pubsub debounce hardening) improves subscription discovery timing. In CI it also surfaced a set of latent shared-log races/timeouts (and a couple of migration/role issues) that could previously hide behind slower peer discovery.

The failure signature that motivated this work:
- `events.spec.ts`: duplicate `replicator:join` (same peer hash appears twice)
- `migration.spec.ts` (migration-8-9): replication never completes within `waitForResolved` default timeout
- `replicate.spec.ts`: timing sensitivity in `waitForReplicator waits until maturity`

This PR makes the pubsub subscription handshake resilient to debounce timing, and makes shared-log replication-info application deterministic (per-peer serialization + retry on startup) so those CI failures stop occurring.

### 1) Pubsub: subscribe debounce window could drop/omit state

- `DirectSub.subscribe()` is debounced; the topic is only fully initialized when `_subscribe()` runs.
- Incoming remote `Subscribe` handling drops the subscription if the topic hasn't been initialized locally yet.
- `requestSubscribers()` overlap responses were based only on `this.subscriptions`, which can be empty during the debounce window.

Net effect: peers can connect during the debounce window and fail to discover each other's subscriptions promptly or at all.

### 2) Shared-log: replication-info handling was racy / lossy

- Replication-info messages could be processed concurrently for the same peer (fire-and-forget async work).
- `addReplicationRange()` emitted `replicator:join` based on a TOCTOU check of index state. Two concurrent handlers could both decide a peer is “new” and emit duplicates.
- During startup, replication-info application could fail with NotStarted/IndexNotStarted and the update would be dropped (no retry), leaving the handshake incomplete and migration replication stuck.
- Startup relied on a snapshot of current pubsub state; it didn’t actively backfill subscriber info.

### 3) Migration + role encoding issues (v8 compat)

- v8 compat path could fail to respond with role info due to missing `await`, a `getRole()` implementation that threw in multi-segment cases, and a role->replication-range conversion that used normalized floats rather than u32 coordinate space.
- `RoleReplicationSegment` offset encoding used the wrong nominator.

### 4) A genuine test flake under heavy load

While re-running the full suite, `u32-simple > sharding > handles peer joining and leaving multiple times` intermittently failed because replica convergence could exceed `waitForResolved`’s default 10s timeout on a loaded machine.

### Pubsub (packages/transport/pubsub/src/index.ts)
- Eagerly initialize per-topic state in `subscribe()` so early remote `Subscribe` messages aren’t dropped.
- Include topics pending in the debounce window when responding to `requestSubscribers()` overlap queries.
- If a pending subscribe is cancelled via `unsubscribe()` before debounce fires, clean up the eagerly created topic state.

### Shared-log (packages/programs/data/shared-log/src/index.ts)
- Serialize replication-info application per peer (per-peer promise chain) so the same sender can’t concurrently run `addReplicationRange()`.
- Track latest replication-info timestamp per peer and ignore older updates.
- If replication-info can’t be applied yet due to NotStarted/IndexNotStarted/Closed conditions, store the latest message per peer and flush after open instead of dropping it.
- After open, call `pubsub.requestSubscribers(this.topic)` and backfill `handleSubscriptionChange(...)` from the current subscriber snapshot.
- Make `replicator:join` idempotent: emit on transition “not known replicator -> has segments”, including restart/full-state announce cases.

### Migration + role fixes
- `getRole()` is now best-effort when multiple local segments exist (selects the widest segment).
- Ensure role responses await `getRole()`.
- Convert role -> replication segments by denormalizing factor/offset into u32 coordinate space.
- Fix `RoleReplicationSegment` offset encoding.

### Tests
- Fix migration test’s v8 mock to reply using the opened instance (`db1.log.rpc`).
- Make `waitForReplicator waits until maturity` assert on remaining maturity time derived from the observed segment timestamp.
- Harden `checkReplicas()` to avoid false negatives under load by using a slightly longer timeout + lower polling rate.

### Debug Docs
- Keep `shared-log-debug.md`, `shared-log-debug-plan.md`, and `scrap.md` as an audit trail for future debugging; these can be removed later once we’re confident.

## Verification

- `pnpm run build` (PASS)
- Targeted tests (PASS):
  - `replicate:join not emitted on update`
  - `8-9, replicates database of 1 entry`
  - `9-8, replicates database of 1 entry`
  - `waitForReplicator waits until maturity`
  - `handles peer joining and leaving multiple times`
- Full CI Part 4 suite (PASS):
  - `pnpm run test:ci:part-4`
  - `@peerbit/shared-log`: `1743 passing`
  - `@peerbit/shared-log-proxy`: `1 passing`

All Checks passed when running on [my repo](https://github.com/Faolain/peerbit/pull/6/checks) with no change TBD on this repo 🤞 

## Commits

- `35963a7ac` tests: checkpoint
- `a2ab6d31d` shared-log: harden replication handshake + reduce flakes
- `bb7e7b031` docs: record shared-log debug commit hashes
- `67125c13f` docs: add shared-log debugging scrap notes

## Additional Notes: 
There is another PR at https://github.com/Faolain/peerbit/pull/7 which the below summarizes vs this PR

Screenshot 2026-02-06 at 5 03 24 PM

### DIFF
--- a/scrap.md
+++ b/scrap.md
@@ -1,0 +1,4208 @@
+
+
+❯ codex --yolo
+╭──────────────────────────────────────────────────────╮
+│ >_ OpenAI Codex (v0.98.0)                            │
+│                                                      │
+│ model:     gpt-5.3-codex xhigh   /model to change    │
+│ directory: ~/Documents/Projects/f/peerbit-pr589/repo │
+╰──────────────────────────────────────────────────────╯
+
+  Tip: New Try the Codex App with 2x rate limits until April 2nd. Run 'codex app' or visit https://chatgpt.com/codex
+
+⚠ `[features].web_search_request` is deprecated. Use `web_search` instead.
+Set `web_search` to `"live"`, `"cached"`, or `"disabled"` at the top level (or under a profile) in config.toml.
+
+/status
+
+╭────────────────────────────────────────────────────────────────────────────────╮
+│  >_ OpenAI Codex (v0.98.0)                                                     │
+│                                                                                │
+│ Visit https://chatgpt.com/codex/settings/usage for up-to-date                  │
+│ information on rate limits and credits                                         │
+│                                                                                │
+│  Model:                gpt-5.3-codex (reasoning xhigh, summaries auto)         │
+│  Directory:            ~/Documents/Projects/f/peerbit-pr589/repo               │
+│  Approval:             never                                                   │
+│  Sandbox:              danger-full-access                                      │
+│  Agents.md:            AGENTS.md                                               │
+│  Account:              <redacted>                                              │
+│  Collaboration mode:   Default                                                 │
+│  Session:              <redacted>                                              │
+│                                                                                │
+│  5h limit:             [████████████████████] 100% left (resets 06:10)         │
+│  Weekly limit:         [███████████░░░░░░░░░] 57% left (resets 17:10 on 9 Feb) │
+╰────────────────────────────────────────────────────────────────────────────────╯
+
+
+› Read shared-log-debug.md and follow the instructions
+
+
+• I need to check the shared-log-debug.md file in the repository. It’s important to carefully follow the instructions provided there. I’ll inspect the file using the terminal commands cat or nl to read and view its content. To do this, I’ll utilize the functions.exec_command. I want to
+  ensure everything is checked thoroughly so that I don’t miss any important details while following the instructions!
+
+• Reading shared-log-debug.md and then I’ll execute whatever debug/build/test steps it asks for.
+
+• Explored
+  └ List ls -la
+
+• I see there's a shared-log-debug.md and a shared-log-debug-plan.md. I need to read shared-log-debug.md first and follow the instructions. It’s pretty long at 56k lines, so maybe I should use less? I’ll try sed -n '1,200p' to check the beginning. Following the instructions may involve
+  running specific commands, so I'll read the whole file to grasp the necessary steps, and I’ll use rg to find sections marked with "##". Let’s inspect!
+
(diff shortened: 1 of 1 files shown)


Faolain:
Reran the failing test and it all passes now

• master already covers this regression; a new PR adding this particular test would be redundant.

  - Direct test is present at replicate.spec.ts:331 (/peerbit/packages/programs/data/shared-log/test/
    replicate.spec.ts:331), including the remaining-maturity timing assertion at replicate.spec.ts:351 (/
    peerbit/packages/programs/data/shared-log/test/replicate.spec.ts:351).
  - Companion coverage is also present at events.spec.ts:15 (/peerbit/packages/programs/data/shared-log/
    test/events.spec.ts:15), events.spec.ts:92 (/peerbit/packages/programs/data/shared-log/test/
    events.spec.ts:92), and migration.spec.ts:124 (/peerbit/packages/programs/data/shared-log/test/
    migration.spec.ts:124).
  - It is in normal CI scope via test:ci:part-4 in package.json (/peerbit/package.json) and executed in
    ci.yml (/peerbit/.github/workflows/ci.yml).
  - Empirical check on current master: this test passed repeatedly (30/30 extra looped runs, plus targeted companion runs).

🚀

PR #673, opened 2026-03-16
Fix unhandled delivery promise rejections in stream

marcus-pousette:
## Summary
- wrap stream delivery promises so late ACK timeout failures are marked handled immediately
- keep the public `publish()` and `publishMessage()` APIs unchanged while routing through private async implementations
- add a regression test that delays awaiting the publish promise until after the delivery timeout

## Problem
`AcknowledgeDelivery` timeouts can reject after the send path has already returned control to the caller. On Node 22, if the caller attaches its handler late or fire-and-forgets the publish, that rejection can surface as an `unhandledRejection` and terminate the process.

## Testing
- `node ./node_modules/aegir/src/index.js run test --roots ./packages/transport/stream -- -t node --grep "does not emit unhandledRejection before callers await delivery timeouts"`
- `node ./node_modules/aegir/src/index.js run test --roots ./packages/transport/stream -- -t node`

### DIFF
--- a/packages/transport/stream/src/index.ts
+++ b/packages/transport/stream/src/index.ts
@@ -2607,7 +2607,18 @@ export abstract class DirectStream<
 	/**
 	 * Publishes messages to all peers
 	 */
-	async publish(
+	public publish(
+		data: Uint8Array | Uint8ArrayList | undefined,
+		options: PublishOptions = {
+			mode: new AcknowledgeAnyWhere({
+				redundancy: DEFAULT_SEEK_MESSAGE_REDUDANCY,
+			}),
+		},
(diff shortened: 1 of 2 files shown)


peerbit-org:
Closing as superseded by the current stream delivery handling on `master`.

The original issue here was unhandled delivery promise rejections when callers attach handlers late or fire-and-forget delivery promises. `master` now marks stream delivery promises as handled internally via `markPromiseHandled`, including the public `publish()` path and internal delivery deferred promises, and has newer unhandled-rejection regression coverage in `stream.spec.ts`.

This PR is now conflicting and the functional fix has already landed, so keeping it open would duplicate/supersede current code.

PR #691, opened 2026-03-24
feat(pubsub): support opt-in parent upgrades

marcus-pousette:
## Summary
- add an opt-in parent-upgrade path for fanout joins so a node can reparent to a materially better direct parent after it already joined
- keep the behavior conservative by requiring explicit configuration rather than changing the default routing policy
- isolate the fanout-tree sim in a child-process runner and add regressions around late direct-edge upgrades

## Why
Today fanout delivery stays correct when a late direct edge appears, but the overlay does not improve its parent choice after it has already attached. This PR adds a guarded path for eventual better-parent convergence without changing the default stability-first behavior.

## Tests
- `pnpm run build`
- `pnpm --filter @peerbit/pubsub test -- -t node`

### DIFF
--- a/packages/programs/data/document/document/test/index.spec.ts
+++ b/packages/programs/data/document/document/test/index.spec.ts
@@ -3502,7 +3502,9 @@ describe("index", () => {
 						expect(writer.docs.index.hasPending).to.be.false;
 					});
 
-					it("onMissedResults respects already emitted results", async () => {
+					it("onMissedResults respects already emitted results", async function () {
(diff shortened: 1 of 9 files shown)


peerbit-org:
Closing as superseded by #911. I checked the branch against #911: the remaining patch-unique work is the older opt-in runtime/teardown slice, which has been replaced by the current guarded parent-upgrade implementation and validation in #911; the document/test cleanup patches are already patch-equivalent upstream.

PR #692, opened 2026-03-26
chore(pubsub): add parent-upgrade evaluation benchmark

marcus-pousette:
## Summary
- add a reusable `fanout-tree-parent-upgrade-eval` benchmark entrypoint for A/B comparison of parent upgrades off vs on
- extend the fanout-tree sim params so the benchmark can pass parent-upgrade options through to the join logic
- add a smaller `live-like` scenario alongside the existing `mild` and `tight` evaluation scenarios

## Why
We are not promoting proactive parent upgrades to a default behavior right now. The useful part of the investigation is the measurement tooling: it lets us compare future routing-policy ideas with the same repeatable command instead of relying on ad hoc shell snippets.

## Verification
- `pnpm --filter @peerbit/pubsub build`
- `pnpm -C packages/transport/pubsub run bench -- fanout-tree-parent-upgrade-eval --scenario mild --seeds 1`

## Notes
This PR does not change the fanout runtime behavior. It only adds benchmark and evaluation support.

### DIFF
--- a/packages/transport/pubsub/benchmark/fanout-tree-parent-upgrade-eval.ts
+++ b/packages/transport/pubsub/benchmark/fanout-tree-parent-upgrade-eval.ts
@@ -4,7 +4,7 @@ import {
 	type FanoutTreeSimResult,
 } from "./fanout-tree-sim-lib.js";
 
-type EvalScenarioName = "mild" | "tight";
+type EvalScenarioName = "mild" | "tight" | "live-like";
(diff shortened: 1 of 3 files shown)


peerbit-org:
Closing as superseded by #911. The parent-upgrade evaluator/benchmark work from this older slice has been folded into the current consolidated fanout parent-upgrade PR, including the updated preset/default handling and CI harness.

PR #705, opened 2026-03-30
fix(blocks): catch LEVEL_DATABASE_NOT_OPEN in put() after stop()

Faolain:
## Summary

- **Bug**: `AnyBlockStore.put()` throws `LEVEL_DATABASE_NOT_OPEN` when a late replication write arrives after the Level-backed blockstore has been closed during shutdown. This surfaces as an unhandled promise rejection, crashing the process or failing tests after assertions pass.
- **Fix**: Catch the `LEVEL_DATABASE_NOT_OPEN` error in `put()` and return the CID as a silent no-op, since the data is being discarded by the closing store.
- **Test**: Added `shutdown-race.spec.ts` that reproduces the bug with a Level-backed store and verifies the fix. (**Note**: this error is only seen with LevelsDB not with the InMemory DB)

## Test plan

- [x] New test fails on unpatched code (verified `LEVEL_DATABASE_NOT_OPEN` throw)
- [x] New test passes with fix applied
- [x] Existing `level.spec.ts` tests still pass

## Full Test Suite Results

Built the entire monorepo (`pnpm run build`) and ran the full `@peerbit/blocks` test suite via aegir on an isolated worktree.

**22 passing, 0 failing**

### New tests (2/2 passing)
- `put() after stop() should not throw LEVEL_DATABASE_NOT_OPEN` — PASS
- `get() after stop() should not throw (baseline check)` — PASS

### Existing tests (20/20 passing)
- **level** (2): `rw`, `iterate`
- **transport** (18): relay reads, provider resolution, timeout, concurrency, iteration, waitForRequest — all pass

No regressions. TypeScript build (`aegir build`) completes cleanly.


marcus-pousette (approved):
Approved after a maintainer polish pass. Verified in a temporary PR worktree with `pnpm run build` and `pnpm --filter @peerbit/blocks test -- -t node`.

PR #706, opened 2026-03-30
fix(indexer-sqlite3): prevent crashes during and after shutdown

Faolain:
## Summary

Four related bugs in `SQLiteIndex` cause crashes during concurrent shutdown:

- **Getters throw after close**: `tables`, `rootTables`, `cursor` threw `NotStartedError` when accessed after `stop()`. Now return empty defaults (empty Map / empty array).
- **Fields uninitialized before init()**: `_rootTables`, `_tables`, `_cursor` were declared with `!` but never initialized in the constructor. Now initialized to empty defaults in field declarations.
- **stop()/drop() crash on undefined fields**: Both methods assumed backing fields were always initialized. Now use optional chaining / null checks.
- **put() throws NotStartedError when closed**: During teardown, late writes threw `NotStartedError` as unhandled rejections. Now returns `undefined` as a silent no-op.

## Test plan

- [x] All 4 new tests fail on unpatched code (verified each failure mode)
- [x] All 4 new tests pass with fixes applied
- [x] Existing `engine.spec.ts` tests still pass

## Full Test Suite Results

Built the entire monorepo (`pnpm run build`) and ran the full `@peerbit/indexer-sqlite3` test suite via aegir on an isolated worktree.

**265 passing, 1 failing** (1 pre-existing failure unrelated to this PR)

### New tests (4/4 passing)
- `getters should not throw after close` — PASS
- `getters should not throw before init` — PASS
- `stop and drop should not crash when fields are uninitialized` — PASS
- `put should return undefined when closed, not throw NotStartedError` — PASS

### Existing tests (261/262)
All existing test suites pass except one pre-existing failure:
- `sort > will not sort by default when fetching all` — FAIL (expects insertion order `["3","2","1"]` but gets `["1","2","3"]`; this test was **not modified** on this branch and fails identically on master — see #704)

No regressions from our changes.

### DIFF
--- a/packages/programs/data/document/document/src/search.ts
+++ b/packages/programs/data/document/document/src/search.ts
@@ -4349,6 +4349,7 @@ export class DocumentIndex<
 					indexBy: this.indexBy,
 					nested: this.nestedProperties,
 				});
+				await index.start();
 				return index;
 			};
 
@@ -4927,6 +4928,7 @@ export class DocumentIndex<
 			indexBy: this.indexBy,
 			nested: this.nestedProperties,
 		});
+		await temporaryIndex.start();
 		for (const value of intoIndexable) {
 			temporaryIndex.put(value);
 		}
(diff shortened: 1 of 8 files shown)


peerbit-org (approved):
Reviewed the lifecycle follow-ups and CI is green after rerunning the unrelated part-7 flake.

PR #707, opened 2026-03-30
fix(document): catch NotStartedError in putWithContext() during shutdown

Faolain:
## Summary

- **Bug**: When two peers replicate a Documents store and one peer closes while a replication write is still in-flight, `putWithContext()` throws `NotStartedError` from `this.index.put()`. This surfaces as an unhandled promise rejection, crashing test runners or applications.
- **Fix**: Wrap `this.index.put()` in `putWithContext()` with a try/catch that returns gracefully on `NotStartedError`. This matches the existing defensive pattern already used in `close()` (line 1442) and `drop()` (line 1469) in the same file.
- **Test**: Added `shutdown-race.spec.ts` with two scenarios: race condition during replication, and clean close after full convergence.

## Test plan

- [x] Fix follows the same `NotStartedError` catch pattern already validated in `close()` and `drop()`
- [x] Test compiles cleanly (TypeScript check passes)
- [x] Full test suite passes via aegir

## Full Test Suite Results

Ran the full `@peerbit/document` test suite using `pnpm --filter @peerbit/document test` (aegir, which properly loads `.mocharc.json` and `chai-as-promised`).

**189 passing, 0 failing**

### New tests (2/2 passing)
- `putWithContext() should not throw NotStartedError when index is stopped during replication` — PASS (866ms)
- `closing a store after replication completes should not throw` — PASS (313ms)

### Existing tests (187/187 passing)
All existing suites pass: borsh, domain, index operations, search, iterate, replication, events, memory, caching, prefetch, migration, custom index, query distribution, count, types, profile, acl.

**No regressions.**


marcus-pousette (approved):
Reviewed the current head after making the shutdown regression deterministic. The production change is narrowly scoped to swallowing NotStartedError only after the document index is closed, and CI is green after rerunning the failing shard.

PR #723, opened 2026-04-13
fix(blocks): retry provider discovery sooner for pending gets

marcus-pousette:
## Summary
- retry provider discovery faster while a remote block get is already waiting and still has no providers
- keep the slower bounded retry cadence once provider candidates exist and actual request retries are in flight
- add a transport regression covering delayed provider discovery during an in-flight get

## Why
File-share open profiling showed the dominant reader-open delay was before nested `Files.open()` started. The gap traced back to `Program.load(...)` waiting on `RemoteBlocks.get(...)` for the root CID, where an empty provider set could sit on a coarse ~5s retry timer before rechecking discovery.

## Validation
- `pnpm run build`
- targeted `@peerbit/blocks` package test entrypoint is currently blocked by an existing aegir/workspace resolution issue in this fresh worktree (`Cannot find module` errors during package-local build), so I validated the change through the full repo build and the new regression coverage added in the transport test file

### DIFF
--- a/packages/transport/blocks/src/remote.ts
+++ b/packages/transport/blocks/src/remote.ts
@@ -524,12 +524,16 @@ export class RemoteBlocks implements IBlocks {
 
 			let requeryCount = 0;
 			const maxRequests = Math.max(1, this.maxRequeryOnReachable);
-			const retryIntervalMs = Math.max(
+			const requestRetryIntervalMs = Math.max(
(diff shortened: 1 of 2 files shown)


peerbit-org:
Superseded by #724 so the PR is owned by the correct peerbit-org GitHub account.

PR #764, opened 2026-05-03
fix(stream): handle internal delivery rejections

peerbit-org:
## Summary
- mark internal delivery promises as handled when they are created
- keep caller-visible publish/publishMessage rejection behavior unchanged
- add a regression test for delivery aborting before publishMessage returns
- ignore local .envrc files

## Verification
- pnpm --filter @peerbit/stream... run build
- pnpm --filter @peerbit/stream test -- --grep "(unhandledRejection|publishMaybe returns false|publishMessageMaybe returns false|will resolve immediately of no neighbours)"
- pnpm --filter @peerbit/stream test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ coverage
 references
 .npmrc
 .env.github
+.envrc
 
 # Supabase local secrets (do not commit)
 .env.supabase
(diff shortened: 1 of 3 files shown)


marcus-pousette (approved):
Reviewed delivery rejection handling fix and regression coverage.

PR #766, opened 2026-05-04
fix(shared-log): cancel stale checked prune on local leadership

marcus-pousette:
## Summary
- cancel stale checked-prune bookkeeping when local leadership is confirmed again
- recheck debounced checked-prune ownership only for hashes that already have active checked-prune work
- add regression coverage for stale debounced ownership with pending delete state

## Notes
- This intentionally excludes diagnostics and benchmark-output changes.
- The branch is based on origin/master from the clean PR worktree.

## Test Plan
- pnpm run build
- NODE_OPTIONS=--max-old-space-size=8192 PEERBIT_TEST_SESSION=mock node ./node_modules/aegir/src/index.js run test --roots ./packages/programs/data/shared-log -- -t node --grep "fetches next blocks once|will prune on put 300 before join|time out when pending IHave|checked prune"
- NODE_OPTIONS=--max-old-space-size=8192 PEERBIT_TEST_SESSION=mock pnpm run test:ci:part-7

### DIFF
--- a/packages/programs/data/shared-log/src/index.ts
+++ b/packages/programs/data/shared-log/src/index.ts
@@ -241,6 +241,12 @@ export { MAX_U32, MAX_U64, type NumberFromType };
 export const logger = loggerFn("peerbit:shared-log");
 const warn = logger.newScope("warn");
 
+type CheckedPruneLeaderMap = Map<string, { intersecting: boolean }>;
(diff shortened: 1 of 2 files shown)


peerbit-org (approved):
CI is green after rerunning part-7. Local verification also passed for build, focused shared-log regression coverage, and part-7 before the type-only cleanup. Scope is limited to the shared-log checked-prune fix and regression tests.

Issue #776, opened 2026-05-06
Investigate production file-share listing latency over public relay


peerbit-org:
## Summary
After dao-xyz/peerbit-examples#139, the 64 MB adaptive production benchmark against https://files.dao.xyz completed successfully, but listing/discovery latency remains unhealthy.

The run passed chunk resolution and download, but discovery/listing lag was 89486 ms.

## Observed run
- Site: https://files.dao.xyz
- Scenario: prod
- Reader role: adaptive
- File size: 64 MB
- Upload: 2975 ms / 180.46 Mbps
- Discovery/listing lag: 89486 ms
- Download: 7346 ms / 73.08 Mbps
- Read-ahead: 2
- Initial local chunks before read: 94/128
- Final listed local chunks: 123/128
- Manifest-head misses: 0

The reader did not display the file until after repeated long root-list attempts:

```text
initial-open: 10017 ms, listCount=0, displayListCount=0
event:        31522 ms, listCount=0, displayListCount=0
refresh:      26241 ms, listCount=0, displayListCount=0
refresh:      26521 ms, listCount=0, displayListCount=0
refresh:      11024 ms, listCount=0, displayListCount=0
refresh:       9410 ms, listCount=1, displayListCount=1
```

## Why this matters
The previous 64 MB failure mode was chunk download correctness. That appears mitigated for this run by reducing remote persisted read-ahead and retrying manifest heads per chunk. The next visible production problem is root-file discovery: the reader has peer hints and eventually receives chunks, but `Documents.index.search(IsNull(parentId))` returns no displayable ready root file for tens of seconds.

## Initial investigation areas
- Document index remote search behavior for `IsNull(parentId)` over public relay / hinted peers.
- Whether root ready-manifest replication/provider announcements lag behind chunk replication.
- Whether search with `throwOnMissing: false` masks partial shard/provider failures as empty results.
- Whether root listing should use a more exact ready-manifest lookup path once a root file event or upload id is known.
- Whether shared-log/document sync prioritization is spending too much work on chunk entries before ready root manifests.

Related: dao-xyz/peerbit#775 and dao-xyz/peerbit-examples#139.

peerbit-org:
Follow-up from the post-#139 production run and code trace:

- The 64 MB production benchmark passed after peerbit-examples#139, but discovery/listing was still 89.486s.
- Reader was connected and had replicator metadata (`replicatorCount=2` before download), but list RPCs repeatedly took 10-31s and returned no root file.
[...]

PR #782, opened 2026-05-06
feat(document): allow foreground reads to cache without blocking

peerbit-org:
## Summary

Adds an opt-in `remote.waitForReplication: false` mode for document queries that request `remote.replicate: true`.

By default, `remote.replicate: true` keeps its existing synchronous contract: search/iterate waits for local replication bookkeeping before resolving. Foreground reads can now choose non-blocking cache-fill behavior: remote results are returned once received, while local replication of those results continues in the background.

This addresses the Peerbit-level contention seen in file-share style workloads where foreground document reads and background chunk replication compete on the same peer/log. See #776.

## What changed

- Added `waitForReplication?: boolean` to `RemoteQueryOptions`.
- Preserved default wait-before-return semantics for existing callers.
- Added non-blocking background replication handling when `waitForReplication: false`.
- Added regression coverage proving both behaviors:
  - default replicated remote search waits for `SharedLog.join` to settle.
  - opt-in foreground search returns results while `SharedLog.join` is still blocked, then local replication settles afterward.

## Validation

- `pnpm run build` passed on rebased `origin/master`.
- `node ./node_modules/aegir/src/index.js run test --roots ./packages/programs/data/document/document -- -t node --grep "search replicate"` passed: 15 passing.
- Before rebasing, `PEERBIT_TEST_SESSION=mock pnpm run test:ci:part-2a` passed: 202 document tests + 3 document-proxy tests.
- Before rebasing, `PEERBIT_TEST_SESSION=mock pnpm run test:ci:part-7` had one intermittent shared-log distribution-objective failure in `u32-simple sharding distribution objectives memory joining half limited`; isolated rerun of `joining half limited` passed for both u32-simple and u64-iblt (`2 passing`).

### DIFF
--- a/packages/programs/data/document/document/src/search.ts
+++ b/packages/programs/data/document/document/src/search.ts
@@ -201,6 +201,13 @@ export type ReachScope = {
 
 export type RemoteQueryOptions<Q, R, D> = RPCRequestAllOptions<Q, R> & {
 	replicate?: boolean;
+	/**
+	 * When `replicate` is true, wait for local replication bookkeeping before
+	 * resolving query results. Defaults to true for existing sync semantics. Set
+	 * to false for foreground reads that should return once remote results arrive
+	 * while caching those results locally in the background.
+	 */
+	waitForReplication?: boolean;
 	minAge?: number;
 	throwOnMissing?: boolean;
 	retryMissingResponses?: boolean;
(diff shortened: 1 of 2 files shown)


Faolain:
Ahhh this is the exact issue I was having!

peerbit-org:
Superseded by #943, which fixes the foreground remote read priority path directly without adding a public background-cache/waitForReplication API.

peerbit-org:
Thanks @Faolain. We moved this to #943, which fixes the underlying foreground remote-read priority path directly and keeps the public API unchanged. Closing this PR in favor of that root fix.

PR #786, opened 2026-05-07
Add Rust-backed indexer package and native query path

peerbit-org:
## Summary

- add `@peerbit/indexer-rust` under `packages/utils/indexer/rust`
- add a Rust/WASM ordered entry store
- add compacted, checksummed snapshots plus an append-only, checksummed put/delete WAL
- add SQLite-like persistence durability modes: default `normal` throughput mode and opt-in `strict` per-write sync mode
- speed up Node persistence with synchronous append writes, a reusable journal file descriptor, and a larger configurable WAL compaction window
- add browser OPFS persistence using `createSyncAccessHandle()` when available, with writable-file fallback
- recover from unclean shutdown by replaying `index.wal`, and recover from torn compaction by falling back to `index.bin.tmp`
- move query planning, scoped filtering, sorting, count, sum, pagination, delete, and batched store lookup into Rust
- route the TypeScript adapter through a native `NativeRustIndex` facade so the store and query engine are owned together in Rust
- reduce hot `put()` bridge overhead with schema-compiled field writers, deterministic numeric field IDs, direct Borsh-compatible byte output, lazy cycle tracking only when needed, and a direct Rust bridge decoder into `DocumentFields`
- avoid transient persistence no-op awaits by keeping the mutation queue on the durable path only
- run the shared indexer conformance suite in both persistent and transient modes
- add a compiled benchmark entrypoint for the shared indexer benchmark suite

This intentionally keeps the public indexer API compatible with the existing TypeScript indexer contract. The Rust package is now a real native indexer core rather than just a storage baseline, and the remaining bridge surface is mostly serialization/API boundary work.

## Verification

- `pnpm --filter @peerbit/indexer-rust build`
- targeted Node put comparison on this machine, flat `{ id: string, number: u32 }` documents, precomputed docs, sequential awaited `put()`:
  - Rust transient, 10,000 puts: `42.63ms` (`~235k ops/s`)
  - SQLite transient, 10,000 puts: `46.18ms` (`~217k ops/s`)
  - Rust persistent default `normal`, 10,000 puts: `113.74ms` (`~88k ops/s`)
  - SQLite persistent `synchronous=NORMAL`, 10,000 puts: `167.45ms` (`~60k ops/s`)
  - Rust transient, 50,000 puts after warmup: `256.13ms` (`~195k ops/s`)
  - SQLite transient, 50,000 puts: `198.09ms` (`~252k ops/s`)
  - Rust persistent default `normal`, 50,000 puts: `579.31ms` (`~86k ops/s`)
  - SQLite persistent `synchronous=NORMAL`, 50,000 puts: `841.62ms` (`~59k ops/s`)
  - Rust persistent `strict`, 1,000 puts: previously measured `5.10s` (`~196 ops/s`), still intentionally fsync-bound

## Benchmark notes

The earlier native-facade benchmark showed concrete query-path wins over the first Rust baseline and adapter stages:

- `number query fetch 10 - transient`: `15,096 ns` -> `10,894 ns` (~28% faster)
- `number query fetch 10 - persist`: `16,591 ns` -> `12,336 ns` (~26% faster)
- `string query matching - transient`: `68,794 ns` -> `61,320 ns` (~11% faster)
- `bool query fetch 10 - transient`: `5,367,347 ns` -> `3,896,186 ns` (~27% faster)
- `non bool query fetch with sort 10 - transient`: `2,050,224 ns` -> `1,954,628 ns` (~5% faster)
- `nested bool query - persist`: `5,300,224 ns` -> `4,022,555 ns` (~24% faster)

Persistent mutation throughput is now above SQLite WAL `synchronous=NORMAL` on the same local put benchmark. The old `~18k ops/s` Rust persistent result was mostly self-inflicted: the Node path used async `fs/promises` writes and compacted the WAL into full snapshots every 1,024 writes. Default compaction is now `64 * 1024` operations, with `persistence.compactAfterOperations` available for tighter replay bounds or tests. Transient 10k puts are now above SQLite in this benchmark. Larger transient batches still expose remaining WASM/bridge and JS-value storage overhead; the next clean improvement is a batched put bridge so the Rust side can absorb many documents per crossing.

## Multi-field planner benchmark

The existing shared benchmark case `multiFieldQueryBenchmark` exposed an inefficient Rust `AND` plan: for each range intersection it materialized both half-range bitmaps before intersecting. The planner now estimates candidate sizes and seeds `AND` from the most selective child, then the final Rust evaluator still enforces the full query. `vector_search` also re-checks the full filter because `AND` candidates are now a deliberate superset.

The shared benchmark fixture now gives every benchmark schema a stable `@variant(...)` name, so SQLite can run the same shared benchmark entrypoint instead of failing during schema setup.

Shared benchmark over 100,000 range docs, OR of four two-field range intersections, fetch 10:

- Rust before this planner fix: transient `~307 ms`, persist `~309 ms`
- Rust after this fix: transient `54.4 us` (`~19.6k ops/s`), persist `55.9 us` (`~19.2k ops/s`)
- Existing simple indexer: transient `163 ms` (`~6 ops/s`)
- SQLite shared benchmark: transient `28.3 ms` (`~35 ops/s`), persist `33.7 ms` (`~30 ops/s`)

So on this find-leaders-style multi-field range intersection, Rust is now roughly `~3,000x` faster than the simple indexer and `~500-600x` faster than SQLite in the shared benchmark run.


Faolain:
What are the drawbacks of swapping the indexer? Would this mean that sqlite native querying capabilities outside of peerbit itself are lost?

marcus-pousette:
@Faolain 

It will be opt-in but fully compatible in terms of API, but the underlying way the storage is represented on disc will change so you would have to launch a new node, sync all the previous data into this new node, to migrate. 

Idea for migration. I will put this as a opt-in implementation (a long with other native work I am working on). And eventually there would be a major bump that moves the default behaviour to use the native path once it is prooved to be stable !

I observed locally some 10-100x performance boosts doing pure native paths of hot-paths for the "shared-log" component and related modules so this is a bit of a high value topic to move forward on

PR #788, opened 2026-05-07
Bridge Rust index planner into indexer adapter

marcus-pousette:
## Summary

Stacked on #787.

This wires the Rust query planner into the `@peerbit/indexer-rust` TypeScript adapter behind a conservative candidate-filtering bridge:

- expose a WASM `NativeQueryPlanner` wrapper for put/delete/clear/query
- index scalar field facts from adapter documents into the Rust planner on put and snapshot load
- use native candidates for supported exact/range predicates, then keep the existing TypeScript evaluator as the residual correctness gate
- fall back safely for unsupported query shapes such as contains/case-insensitive string matching, nested queries, and mixed unsupported OR/NOT branches
- index `Uint8Array` fields both as exact byte blobs and as per-byte numeric facts so existing conformance behavior is preserved

## Benchmark signal

The bridge shows the expected wins on selective scalar predicates while broad/unsupported shapes still scan through the existing TS evaluator.

Transient examples from `pnpm --filter @peerbit/indexer-rust benchmark`:

- `number query fetch 10`: ~71.8k ops/s
- `string query matching`: ~6.6k ops/s
- `string count no-matches`: ~539k ops/s
- broad 3-field OR query: still ~6 ops/s because it safely falls back

Persistent examples:

- `number query fetch 10`: ~65.3k ops/s
- `string query matching`: ~6.6k ops/s
- `string count no-matches`: ~533k ops/s

## Verification

- `cargo test`

### DIFF
--- a/packages/utils/indexer/rust/Cargo.lock
+++ b/packages/utils/indexer/rust/Cargo.lock
@@ -73,6 +73,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "itoa"
+version = "1.0.18"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8f42a60cbdf9a97f5d2305f08a87dc4e09308d1276d28c869c684d7777685682"
+
+[[package]]
 name = "js-sys"
 version = "0.3.98"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


peerbit-org:

PR #789, opened 2026-05-07
Use Borsh binary bridge for Rust planner adapter

marcus-pousette:
## Summary

Stacked on #788.

This replaces the temporary JSON bridge between the TypeScript adapter and the Rust planner with a compact Borsh-compatible binary ABI:

- TS encodes planner facts/query payloads with `@dao-xyz/borsh` `BinaryWriter`
- Rust decodes DTOs with `borsh-rs` `BorshDeserialize`
- bridge payloads include a version byte for future ABI changes
- removes `serde`/`serde_json` from the adapter bridge path
- keeps the same residual TypeScript correctness gate from #788

The enum tag order is documented on both sides because that order is part of the bridge ABI.

## Benchmark signal

Compared with the JSON bridge from #788, this is modestly faster on the paths where bridge overhead matters most:

- transient `number query fetch 10`: ~76.7k ops/s, up from ~71.8k
- persistent `number query fetch 10`: ~68.1k ops/s, up from ~65.3k
- transient `string count no-matches`: ~691.7k ops/s, up from ~539k
- persistent `string count no-matches`: ~667.0k ops/s, up from ~533k

Exact string matching is roughly neutral at ~6.5k-6.6k ops/s, which suggests that candidate materialization/residual checking dominates there now.

## Verification

- `cargo test`

### DIFF
--- a/packages/utils/indexer/rust/Cargo.lock
+++ b/packages/utils/indexer/rust/Cargo.lock
@@ -3,6 +3,30 @@
 version = 4
 
 [[package]]
+name = "borsh"
+version = "1.6.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cfd1e3f8955a5d7de9fab72fc8373fade9fb8a703968cb200ae3dc6cf08e185a"
+dependencies = [
+ "borsh-derive",
+ "bytes",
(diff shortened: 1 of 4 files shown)


peerbit-org:

PR #790, opened 2026-05-07
Page native Rust index results

marcus-pousette:
## Summary

Stacked on #789.

This adds native result modes on top of the Borsh bridge:

- add `NativeQueryIndex::count` and `search_page`
- expose WASM `NativeQueryPlanner.count` and `query_page`
- use native count for exact single-field predicates
- use native paged iteration for exact unsorted single-field predicates, so `next(n)` only moves the requested page of IDs across the WASM boundary
- keep compound queries on the residual TypeScript correctness path, because flattened native facts can be a superset for array/nested AND semantics

## Correctness note

The first implementation tried to treat all compiled-native `And` queries as exact. The conformance suite caught why that is unsafe: flattened array facts do not prove that multiple predicates hit the same nested array element. This PR therefore only skips residual evaluation for a single supported field predicate. Compound queries still benefit from native candidate narrowing, but they are checked by the existing TS evaluator.

## Benchmark signal

Compared with #789:

- transient `number query fetch 10`: ~93.9k ops/s, up from ~76.7k
- persistent `number query fetch 10`: ~80.6k ops/s, up from ~68.1k
- transient `string query matching`: ~76.9k ops/s, up from ~6.6k
- persistent `string query matching`: ~73.6k ops/s, up from ~6.5k
- transient `string count matching`: ~1.09M ops/s
- persistent `string count matching`: ~1.08M ops/s
- broad multi-field OR remains ~6 ops/s because it correctly stays on the residual scan path

## Verification

- `cargo test`

### DIFF
--- a/packages/utils/indexer/rust/src/index.ts
+++ b/packages/utils/indexer/rust/src/index.ts
@@ -22,6 +22,13 @@ type NativeQueryPlanner = {
 	clear: () => void;
 	len: () => number;
 	query: (query: Uint8Array, sort: Uint8Array) => string[];
+	query_page: (
+		query: Uint8Array,
+		sort: Uint8Array,
+		offset: number,
+		limit: number,
+	) => string[];
+	count: (query: Uint8Array) => number;
 };
 
 type WasmModule = {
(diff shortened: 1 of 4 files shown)


peerbit-org:

PR #791, opened 2026-05-07
Evaluate scoped native index queries

marcus-pousette:
Stacked on #790.

## What changed
- Adds scoped native field facts so array/object-array predicates can be evaluated against the same element in Rust.
- Moves supported field/logical query correctness into the Rust planner: integer ranges, exact values, bool/bytes, string exact/prefix/contains with case-insensitive support, `is_null`, `and`, `or`, and `not`.
- Removes TS residual filtering for native-supported query ASTs; unsupported query shapes still fall back before compiling a partial native plan.
- Makes native count/page/query use the Rust evaluator, not just candidate bitmaps.
- Adds bridge coverage proving compound array predicates do not match across different array elements.

## Verification
- `cargo test` in `packages/utils/indexer/rust`
- `pnpm --filter @peerbit/indexer-rust test` (node, browser, webworker)

## Benchmark notes from this run
- Transient number query fetch 10: ~13.6us avg, ~76.9k ops/s avg.
- Persist number query fetch 10: ~15.2us avg, ~69.3k ops/s avg.
- Transient string query matching: ~63.9us avg, ~16.1k ops/s avg.
- Persist string query matching: ~64.9us avg, ~15.9k ops/s avg.
- Broad 3-field OR small fetch remains the slow path at ~313ms avg in both transient and persist runs.

### DIFF
--- a/packages/utils/indexer/rust/src/index.ts
+++ b/packages/utils/indexer/rust/src/index.ts
@@ -45,6 +45,7 @@ type NativeValue =
 	| { type: "string"; value: string };
 
 type NativeFieldFact = {
+	scope: number;
 	field: string;
 	value: NativeValue;
 };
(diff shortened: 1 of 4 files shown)


peerbit-org:

PR #793, opened 2026-05-07
Page sorted native index results

peerbit-org:
## Summary
- bridge scalar sort fields into the native Rust planner page path
- allow native page plans for unfiltered `All` queries, so sorted full scans can stay in Rust
- add first-scalar ordered sort indexes for bool, i64, u64, string, and bytes values
- add Rust and TS bridge coverage for native sorted paging, descending order, missing values, mixed scalar ranks, and first-scalar array semantics

Stacked on #791.

## Benchmark
Final run: `pnpm --filter @peerbit/indexer-rust benchmark`

Key row versus the previous stacked branch baseline:
- `non bool query fetch with sort 10 - transient`: `1,777,742 ns` avg, down from ~`22,583,873 ns` avg, about `12.7x` faster
- `non bool query fetch with sort 10 - persist`: `1,955,429 ns` avg, down from ~`23,119,094 ns` avg, about `11.8x` faster

Tradeoff: put paths now maintain first-scalar sort indexes. The final benchmark showed higher and noisier put averages on persistent storage, while medians remained low; this PR optimizes the repeated sorted-page read path rather than write-heavy workloads.

## Validation
- `cargo fmt`
- `cargo test`

### DIFF
--- a/packages/utils/indexer/rust/src/index.ts
+++ b/packages/utils/indexer/rust/src/index.ts
@@ -76,8 +76,14 @@ type NativeQueryCompileResult = {
 	exact: boolean;
 };
 
+type NativeSortField = {
+	field: string;
+	direction: "asc" | "desc";
+};
+
 type NativeCandidatePage = {
 	compiled: NativeQueryCompileResult;
+	sort: NativeSortField[];
 	offset: number;
 	limit?: number;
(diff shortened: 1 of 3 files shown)


peerbit-org:

PR #794, opened 2026-05-07
Remove rust indexer query fallbacks

peerbit-org:
## Summary
- remove the TypeScript scan/query evaluator fallback from `@peerbit/indexer-rust`
- route iterate/count/sum/delete-by-query through native Rust planner operations
- compile explicit `Nested` queries into native prefixed field paths guarded by array-element markers
- add native Rust planner sum and delete_matching operations
- add bridge tests proving the fallback evaluator methods are gone, explicit nested queries run natively, and sum/delete use native queries

Stacked on #793.

## Benchmark
Final run: `pnpm --filter @peerbit/indexer-rust benchmark`

Key rows:
- `non bool query fetch with sort 10 - transient`: `2,317,683 ns` avg
- `non bool query fetch with sort 10 - persist`: `2,143,780 ns` avg

This remains roughly 10x faster than the pre-native-sort baseline (~22-23ms). It is slightly slower than the #793 run (~1.8-2.0ms), likely from run variance / wasm shape changes; the important change in this PR is removing the behavioral TS fallback, not adding a new sorted-read optimization.

## Validation
- `cargo fmt`
- `cargo test`

### DIFF
--- a/packages/utils/indexer/rust/src/index.ts
+++ b/packages/utils/indexer/rust/src/index.ts
@@ -1,11 +1,6 @@
 import { BinaryWriter, deserialize, serialize } from "@dao-xyz/borsh";
 import * as types from "@peerbit/indexer-interface";
-import { logger as loggerFn } from "@peerbit/logger";
-import { equals } from "uint8arrays";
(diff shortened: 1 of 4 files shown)


peerbit-org:

PR #796, opened 2026-05-07
Batch rust indexer store result lookups

peerbit-org:
## Summary

Stacked on #794.

This reduces the remaining adapter overhead in the Rust indexer result path by batching store row materialization:

- adds native `NativeIndexStore.get_many` for query result hydration
- adds native `NativeIndexStore.delete_many` for query deletes
- replaces TS per-key `store.get()` / `store.delete()` loops with one wasm call per result page/delete result set

## Why

After moving query evaluation, sorting, sum, and delete matching into Rust, the next hot path was the wrapper crossing the wasm boundary once per returned store key. For fetch-heavy shared-log/indexer cases, this keeps planner work native and also hydrates returned entries in bulk.

## Validation

- `cargo fmt && cargo test`

Benchmark highlights from this branch:

- `non bool query fetch with sort 10 - transient`: `2,050,224 ns` avg
- `non bool query fetch with sort 10 - persist`: `1,837,264 ns` avg

Compared with the previous stacked PR run (`2,317,683 ns` transient, `2,143,780 ns` persist), this is about 12-14% faster on the existing sorted-fetch benchmark, while keeping the change scoped to result materialization.

### DIFF
--- a/packages/utils/indexer/rust/src/index.ts
+++ b/packages/utils/indexer/rust/src/index.ts
@@ -5,7 +5,9 @@ import { createSnapshotFile, type SnapshotFile } from "./persistence.js";
 type NativeIndexStore<T extends Record<string, any>> = {
 	put: (key: string, id: types.IdKey, value: T) => void;
 	get: (key: string) => [types.IdKey, T] | undefined;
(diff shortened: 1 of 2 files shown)


peerbit-org:

PR #797, opened 2026-05-07
fix(shared-log): prioritize foreground traffic over sync

peerbit-org:
Stacked on #795.

This demotes background shared-log sync traffic to the default background priority while leaving document foreground remote queries above it. The scheduler can still use all bandwidth when no higher-priority work exists, but query/collect traffic no longer sits behind rateless/simple repair messages under cold-start backfill pressure.

I ran:
- pnpm --filter @peerbit/shared-log run build
- pnpm --filter @peerbit/shared-log test -- --no-build --grep "sync-(chunking|repair-session|priority)"
- shared-log half of pnpm test:ci:part-6: 1516 passing; local proxy subcommand fails because @peerbit/shared-log-proxy has no node test files in this worktree

### DIFF
--- a/packages/programs/data/shared-log/src/index.ts
+++ b/packages/programs/data/shared-log/src/index.ts
@@ -175,7 +175,11 @@ import type {
 	Syncronizer,
 } from "./sync/index.js";
 import { RatelessIBLTSynchronizer } from "./sync/rateless-iblt.js";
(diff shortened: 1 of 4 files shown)


Faolain:
Ahhh this explains the issue I was seeing!

PR #798, opened 2026-05-07
Route rust indexer through native facade

peerbit-org:
## Summary

Stacked on #796.

This adds a `NativeRustIndex` facade that owns both the native store and native query planner, then routes the Peerbit TS wrapper through that single object.

The Peerbit index path now uses one native object for:

- snapshot restore indexing
- `put`
- query result materialization, including sorted pages
- query deletes
- `count` / `sum`
- snapshot enumeration / size / clear

The older exported `NativeIndexStore` and `NativeQueryPlanner` remain available for now, but the actual `RustIndex` wrapper no longer shuttles planner keys through JS to another native object.

## Why

After #796, result row lookup was batched, but TS still had to call planner, receive store keys in JS, and then call back into the native store. This PR removes that split for the production wrapper so query planning and row materialization stay inside Rust for each operation.

## Validation

- `cargo fmt && cargo test`

Benchmark highlights versus #796 on the same existing suite:

- `number query fetch 10 - transient`: `15,096 ns` -> `10,894 ns` (~28% faster)
- `number query fetch 10 - persist`: `16,591 ns` -> `12,336 ns` (~26% faster)
- `string query matching - transient`: `68,794 ns` -> `61,320 ns` (~11% faster)
- `bool query fetch 10 - transient`: `5,367,347 ns` -> `3,896,186 ns` (~27% faster)
- `non bool query fetch with sort 10 - transient`: `2,050,224 ns` -> `1,954,628 ns` (~5% faster)
- `nested bool query - persist`: `5,300,224 ns` -> `4,022,555 ns` (~24% faster)

Some persistent bool/sorted rows were roughly flat to slightly slower in this run, so the practical claim here is narrower: the facade removes the avoidable JS/native object handoff and improves several fetch-heavy rows, while keeping conformance green across Node/browser/webworker.

### DIFF
--- a/packages/utils/indexer/rust/src/index.ts
+++ b/packages/utils/indexer/rust/src/index.ts
@@ -2,39 +2,36 @@ import { BinaryWriter, deserialize, serialize } from "@dao-xyz/borsh";
 import * as types from "@peerbit/indexer-interface";
 import { createSnapshotFile, type SnapshotFile } from "./persistence.js";
 
-type NativeIndexStore<T extends Record<string, any>> = {
-	put: (key: string, id: types.IdKey, value: T) => void;
+type NativeRustIndex<T extends Record<string, any>> = {
+	put: (
+		key: string,
+		id: types.IdKey,
+		value: T,
+		fields: Uint8Array,
+	) => void;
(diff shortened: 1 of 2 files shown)


peerbit-org:

PR #952, opened 2026-06-05
fix(shared-log): scope repair sweeps to affected ranges

peerbit-org:
## Summary
- keeps shared-log repair sweeps range-scoped only for join-warmup candidates, where pending peer ranges are a safe bound
- forces full repair sweeps for churn and authoritative join repair because those pending ranges can be incomplete or transient during rebalance
- falls back to a full sweep when pending ranges have no positive-width coverage
- removes the per-sweep global candidate hash set so full scans do not retain every candidate hash in memory
- treats replaced replication ranges as churn for immediate repair delivery, so shrink/shift updates do not trust stale gid-level history before pruning the old owner
- flushes local replication-change work when an awaited `replicate()` call removes or replaces one of this peer's ranges
- stops adaptive replication timers/controllers when a peer switches to fixed/range replication, preventing stale adaptive ranges from being re-added beside fixed ranges
- accepts checked-prune confirmations from current remote leaders while still relying on delete-time ownership revalidation before local removal
- lets `RequestIPrune` responders confirm when they are current `roleAge: 0` leaders, avoiding prune stalls while late-join roles are still maturing under load
- restores `slowDownSend` test monkeypatches and trims two heavy synthetic range-replacement cases so they exercise the same path without dominating shared-log CI shards

## Why
The first #952 direction was too aggressive: using pending peer ranges for every repair mode can miss entries during churn or authoritative join because those ranges are not a complete candidate bound while membership is moving. Warmup still benefits from range-scoped scans, but correctness-critical repair modes now use the original full-scan behavior.

The latest `part-7d` failure showed another replacement-specific bug. Shrinks and shifts are encoded as `replaced` ranges, but the fresh-delivery path only handled `removed` ranges. That let the old owner rely on stale gid-level history and skip sending reassigned entries to the new per-entry leaders before checked pruning. Replaced ranges now use the same fresh-delivery/churn path as removals.

A follow-up `part-7a` failure showed that switching from default adaptive replication to fixed replication did not stop the adaptive rebalance loop. The old adaptive controller could re-add a dynamic range next to the fixed range, causing over-replication in the replication index. Fixed/range replication now explicitly stops adaptive timers and controller state first.

The CI failures also exposed checked-prune liveness issues. A requester can decide an entry is prunable using the current leader view, while the remote responder previously waited on the mature-replicator view before confirming. Under late joins and CI load that could leave entries permanently pending-prunable even though a current leader already had the block. The responder now uses the same current-leader view for confirmation, and local deletion still revalidates ownership before removing the block.

## Validation
- `git diff --check`
- `pnpm --filter @peerbit/shared-log run build`
- `node ./node_modules/aegir/src/index.js run test --roots ./packages/programs/data/shared-log -- -t node --no-build --grep "u32-simple replication references joins by references"`
- `PEERBIT_MOCHA_LOG_TEST_START=1 PEERBIT_MOCHA_LOG_INTERVAL_MS=10000 pnpm run test:ci:part-7a`
- `node ./node_modules/aegir/src/index.js run test --roots ./packages/programs/data/shared-log -- -t node --no-build --grep "u64-iblt sharding (scopes warmup repair sweep candidate scans to pending peer ranges|falls back to a full repair sweep when pending ranges have no coverage|uses a full repair sweep for churn candidates)"`
- `PEERBIT_MOCHA_LOG_TEST_START=1 PEERBIT_MOCHA_LOG_INTERVAL_MS=10000 node ./node_modules/aegir/src/index.js run test --roots ./packages/programs/data/shared-log -- -t node --no-build --grep "u64-iblt (redundancy only sends entries once, 3 peers|sharding (survives deterministic delayed join and leave churn|survives deterministic pubsub join and leave churn|scopes warmup repair sweep candidate scans to pending peer ranges|falls back to a full repair sweep when pending ranges have no coverage|uses a full repair sweep for churn candidates))"`
- `PEERBIT_MOCHA_LOG_TEST_START=1 PEERBIT_MOCHA_LOG_INTERVAL_MS=10000 node ./node_modules/aegir/src/index.js run test --roots ./packages/programs/data/shared-log -- -t node --no-build --grep "u64-iblt replication degree update replace range with another node write (before join with slowed down send|after join)"`
- `PEERBIT_MOCHA_LOG_TEST_START=1 PEERBIT_MOCHA_LOG_INTERVAL_MS=10000 node ./node_modules/aegir/src/index.js run test --roots ./packages/programs/data/shared-log -- -t node --no-build --grep "^u64-iblt replication degree"`
- `PEERBIT_MOCHA_LOG_TEST_START=1 PEERBIT_MOCHA_LOG_INTERVAL_MS=10000 node ./node_modules/aegir/src/index.js run test --roots ./packages/programs/data/shared-log -- -t node --no-build --grep "^u32-simple sharding(?! distribution objectives memory)"`
- `pnpm run test:ci:part-7b`
- `pnpm run test:ci:part-7c`
- `PEERBIT_MOCHA_LOG_TEST_START=1 PEERBIT_MOCHA_LOG_INTERVAL_MS=10000 pnpm run test:ci:part-7d`

### DIFF
--- a/package.json
+++ b/package.json
@@ -102,7 +102,7 @@
 		"test:ci:part-7:cov": "aegir run test:cov --roots ./packages/programs/data/shared-log -- --no-build --grep \"^(u32-simple|u64-iblt)\"",
 		"test:shared-log:chaos": "aegir run test --roots ./packages/programs/data/shared-log -- -t node --no-build --grep \"(control per commmit put before join converges under deterministic (delayed repair traffic|pubsub chaos)|survives deterministic delayed join and leave churn)\"",
 		"test:shared-log:chaos:all": "aegir run test --roots ./packages/programs/data/shared-log -- -t node --no-build --grep \"deterministic\"",
-		"test:shared-log:sharding-regression": "aegir run test --roots ./packages/programs/data/shared-log -- -t node --no-build --grep \"reproduces bounded prune convergence under delayed join traffic\"",
+		"test:shared-log:sharding-regression": "aegir run test --roots ./packages/programs/data/shared-log -- -t node --no-build --grep \"converges under delayed join traffic\"",
 		"coverage": "aegir run coverage",
 		"build": "aegir run build",
 		"clean": "aegir run clean",
--- a/packages/log/src/entry-index.ts
+++ b/packages/log/src/entry-index.ts
@@ -58,11 +58,7 @@ export type NativeLogGraph = {
 	clear: () => void;
 	heads: (gid?: string) => string[];
 	countHasNext: (next: string, excludeHash?: string) => number;
-	shadowedGids: (
-		gid: string,
-		next: string[],
-		excludeHash?: string,
-	) => string[];
+	shadowedGids: (gid: string, next: string[], excludeHash?: string) => string[];
 };
 
 type ResolveFullyOptions =
@@ -326,10 +322,9 @@ export class EntryIndex<T> {
 				const resolved = await Promise.all(
 					hashes.map((hash) => this.resolve(hash, resolveInFullOptions)),
 				);
-				return resolved.filter((entry) => !!entry) as ReturnTypeFromResolveOptions<
-					R,
-					T
-				>[];
+				return resolved.filter(
+					(entry) => !!entry,
+				) as ReturnTypeFromResolveOptions<R, T>[];
 			}
 
 			const shallow = await Promise.all(
@@ -380,7 +375,9 @@ export class EntryIndex<T> {
 			? ({ hash: true } as const)
 			: ((options as { shape: Shape })?.shape as Shape);
 
-		let iteratorRef: ReturnType<typeof this.properties.index.iterate> | undefined;
+		let iteratorRef:
+			| ReturnType<typeof this.properties.index.iterate>
+			| undefined;
 		let iteratorPromise:
 			| Promise<ReturnType<typeof this.properties.index.iterate>>
 			| undefined;
@@ -514,6 +511,7 @@ export class EntryIndex<T> {
 
 	async hasMany(hashes: Iterable<string>) {
 		const batchSize = 64;
+		const directLookupThreshold = batchSize;
 		const existing = new Set<string>();
 		const missing: string[] = [];
 		for (const hash of new Set([...hashes].filter(Boolean))) {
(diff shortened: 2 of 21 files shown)


peerbit-org:
## Investigation findings

A full root-cause investigation of this branch's red CI was run on 2026-06-11 (local repro + git bisect + instrumented runs + per-subsystem diff analysis). Summary:

**The CI failures are branch-introduced regressions, not mainline flakes.** Master passes all of the failing tests locally (9/9 matched tests, `distribute` 6/6 repeat runs); this branch fails 5 of the same 9 deterministically on the same machine/toolchain.

### Proven root causes

1. **`onlyIntersecting: true` prune-leader queries** (introduced in `3b9c70924`, pinned by `git bisect run`): three prune paths now only count peers whose range *strictly contains* an entry's coordinate as leaders, skipping the closest-mature-range fallback in `getSamples`. Leader sets can be empty or exclude the local peer → checked-prune cancellation ("Failed to delete, is leader again") never fires and prune liveness stalls during factor transitions. A 4-line revert fixes `cancels pending checked prune…`, `to smaller then to larger`, `2 peers`, and `only sends entries once` locally (2/2 runs).

2. **False "I have this" answers permanently disable repair** (the `distribute` stall at exact multiples of 64): `sync/simple.ts` now answers possession from `log.blocks.has` (block store) instead of `log.has` (joined log), and confirms-on-offer via `ConfirmEntriesMessage`. Hash-level tracing showed the joiner confirming **all 300 hashes while holding 128–256 entries**; senders mark everything known-by-joiner, `sendFullCoverageJoinRepairEntriesNow` then sends `queuedTotal=0` every pass, and the joiner freezes at k×64 (`REPAIR_EXCHANGE_HEADS_MAX_HEADS`).

Applying both fixes together still leaves `distribute`/`to smaller then to larger` flaky — the rewritten repair pipeline has further independent broken legs (ACK-aborted 64-head batches without retry, prune-of-fresh-entries on the receive path, tautological `acceptsFullReplicaRepairBatch`, removed `minReplicas` append fan-out guard, un-deduplicated `scheduleCurrentHeadsPruneRetry` timer storms, packages/log fast-path `_length` inflation via `unique: true` puts). The branch's test relaxations (participation bounds 0.4–0.6 → 0.25–0.8, call budgets ×2, retry-loop asserts) still fail in CI, i.e. they compensated for regressions rather than fixing them.

### Recommendation

Close this PR rather than iterating on it. The independently safe pieces have been extracted as standalone PRs from `master`:

- #962 — `fix(shared-log): close range iterators deterministically`
- #963 — `fix(stream): clear anySignal listeners after outbound stream creation`
- #964 — `fix(shared-log): stabilize PID control for memory-limited replicators`

The legitimate kernel of this PR — range-scoping which entries the warmup repair sweep *scans* — can be re-introduced from master afterwards (tracked in #961), provided leader-set semantics (closest-leader fallback) are left untouched.

peerbit-org:
Closing per the investigation findings above: the independently safe pieces landed via #962, #963 and #964, and the range-scoped sweep optimization is tracked in #961. The branch is kept for reference.

PR #968, opened 2026-06-12
fix(indexer-sqlite3): settle colliding index-creation scopes

### LINKED ISSUE #967: shared-log: u64-iblt churn leaves entries under-replicated with no retry (seed 13331)
## Summary

`u64-iblt sharding survives deterministic delayed join and leave churn` (chaos suite) fails on CI hardware with entries settling at **1 replica where 2 are required**. Instrumented evidence shows this is a **silent bulk delivery/repair gap in the rateless-IBLT sync path**, not a wrong prune.

## Evidence (chaos seed 13331, CI ubuntu-22.04 runners)

- Nightly run 27350119629 and seed-sweep run 27352602627: same failure signature (different entry hashes per schedule). A second failing seed (24985) shows the identical signature. The u32-simple variant of the same test has never failed in any sweep — only u64-iblt.
- Seed hit-rate: 2/2 un-instrumented, 1/12 instrumented (run 27354526246) — timing-window dependent; never reproduces locally on fast hardware (0/3).
- The instrumented failing run (artifact `flake-hunt-27354526246`, `iter-10-seed-13331.log`) shows:
  - **Zero prune executions** for the whole run (traces at both checked/unchecked removal points never fired) — no entry was ever deleted.
  - **20 of ~36 entries** held ONLY by db0 (the writer). db1 ends with 6 entries, db2 with 22.
  - At assertion timeout the system is fully idle: `syncInFlight` empty on all peers, no pending deletes — everyone believes convergence is reached, with no retry scheduled.

## Interpretation

After delayed join/leave churn, the IBLT reconciliation between db0 and the under-filled peers either never runs for the affected ranges, decodes "no difference" incorrectly, or loses its announce/delivery round under CPU starvation — and nothing re-attempts afterwards. The missing entries are invisible to the requesting side (they never learn the hashes exist), so the `syncInFlightQueue` retry machinery never engages.

## Reproduction

Branch `ci/shared-log-flake-hunt` contains a dispatch/push-triggered workflow that loops the test with `PEERBIT_SHARED_LOG_CHAOS_SEED=13331` and uploads failing logs (post-mortem dump of holders/segments/sync state + prune traces). Expect roughly 1 failure per ~10 iterations on CI runners.

## Next steps

- Trace the IBLT session lifecycle (session open/close, symbol rounds, decode outcome, RequestAll fallback, `onMaybeMissingEntries` announce counts) in a failing run to see whether db0↔db1 sessions for the affected ranges ever start after the last churn event.
- Review what re-triggers a sync session when a previous one is lost mid-flight (peer disconnect during churn) — the idle-with-missing-entries end state suggests session loss is permanent until some unrelated event.

## Context

Found while re-baselining the chaos suite after #965 un-broke the nightly build (the suite had not run for days). Related: #966 (fanout sim wall), #961 (repair scan-scoping design).

peerbit-org:
## Summary

Fixes a silent, permanent, per-process hang in the SQLite query planner that is the root cause of #967 (shared-log entries left under-replicated after churn in the u64 domain).

When two query-plan keys with the same column set race their **first** execution, the losing scope's `beforePrepare` hits the `pendingIndexCreation` collision branch and returns without ever settling its own deferred. That index candidate then stays `created() === false` for the lifetime of the process, and **every later query for that plan key awaits the orphaned promise forever** — a pending await, no rejection, zero CPU, event loop healthy.

In the u64 domain the colliding pair exists by construction: `reduceResolution` nullifies values to `0` or `HALF_MAX`, so the same structural query (e.g. the `getClosest` walk in shared-log's `getSamples`) produces **two plan keys sharing one index key**. Whichever peer loses the startup race on a slow runner wedges; its `_findLeaders` never resolves for affected coordinates, so it drops every entry delivered to it as "not leader" while re-requesting the same entries forever.

The fix settles the colliding command's deferred from the in-flight creation (`create index if not exists` makes the other scope's success equivalent), propagates failures, and observes deferred rejections in the `created()` side channel so they cannot become unhandled rejections.

## Evidence trail (#967)

- 13 instrumented CI rounds on the chaos seed branch narrowed the failure: no prunes ever fire → the starved peer requests the missing hashes every retry window and the holder responds → the starved peer drops the arriving entries as "not leader" thousands of times → its composed `_findLeaders` never resolves (5 s/8 s probes) while **every dependency answers in 0–4 ms** (the probes bypass the planner; the wedged path goes through it).
- Static + dynamic analysis isolated the orphaned-deferred branch; a scripted repro against the shipped dist showed the poisoned plan key timing out minutes later while healthy keys resolve, and resolving with this one-branch fix.

## Testing

- New red-green regression test (`query-planner.spec.ts`): two plan keys with a shared index key and a gated `exec` stub — fails with `colliding scope never settled` on the unfixed code, passes with the fix; also asserts fresh scopes for the previously-collided key settle and the index is created exactly once.
- Full `@peerbit/indexer-sqlite3` suite: 290 passing.
- System-level validation on the #967 hunt branch (chaos seed 13331 × 20 on CI runners, baseline 4–7 failures per round) is running and will be posted to #967.

Fixes #967 (pending the system-level validation round).

### DIFF
--- a/packages/utils/indexer/sqlite3/src/query-planner.ts
+++ b/packages/utils/indexer/sqlite3/src/query-planner.ts
@@ -251,9 +251,21 @@ export class QueryPlanner {
 				if (indexCreateCommands != null) {
 					const commandsToCreate: typeof indexCreateCommands = [];
 					for (const command of indexCreateCommands) {
-						if (this.pendingIndexCreation.has(command.key)) {
-							// TODO is this kind of debouncing needed? how do we end up here?
-							await this.pendingIndexCreation.get(command.key);
+						const pending = this.pendingIndexCreation.get(command.key);
+						if (pending) {
+							// Another plan key with the same column set is already creating
+							// this index ("create index if not exists ..."). Settle THIS
+							// scope's deferred from the in-flight creation: leaving it
+							// pending keeps created() === false for the lifetime of the
+							// process and every later query for this plan key awaits a
+							// promise nobody will ever resolve.
+							try {
+								await pending;
+								command.deferred.resolve();
+							} catch (error) {
+								command.deferred.reject(error);
+								throw error;
+							}
 							continue;
 						}
 						commandsToCreate.push(command);
@@ -323,9 +335,16 @@ export class QueryPlanner {
 						});
 
 						let created = false;
-						deferred.promise.then(() => {
-							created = true;
-						});
+						deferred.promise.then(
+							() => {
+								created = true;
+							},
+							() => {
+								// Rejections are observed by the scopes awaiting this
+								// creation; without this handler a rejected creation turns
+								// into an unhandled rejection through this side channel.
+							},
+						);
 						indexStats.results.push({
 							used: 0,
 							times: [],
--- a/packages/utils/indexer/sqlite3/test/query-planner.spec.ts
+++ b/packages/utils/indexer/sqlite3/test/query-planner.spec.ts
@@ -16,6 +16,67 @@ import {
 	flattenQuery,
 } from "../src/query-planner.js";
 
+describe("QueryPlanner", () => {
+	it("settles scopes whose index creation collides across plan keys", async () => {
+		// Two plan keys (values nullified to different anchors) that share one
+		// index key (same table + columns). The scope that loses the creation
+		// race must still settle its own deferred, otherwise every later query
+		// for its plan key awaits a promise nobody resolves (#967).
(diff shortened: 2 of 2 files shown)


peerbit-org:
System-level validation complete: chaos seed 13331 × 20 on CI runners = **0/20 failures** with this fix (run 27399104355), against a 4–7/20 failure baseline across five identical prior rounds. Details in #967.

PR #988, opened 2026-07-03
Native network plane: rust-core protocols, fused receive, and the pure-native path

peerbit-org:
The complete native network plane for `@peerbit/network-rust` — stacked on #939 (base: `feat/native-log-append-kernel`). This grew from the wire-codec bootstrap slice into the full arc: once bytes arrive from a socket, Rust processes messages end to end — envelope decode+verify, routing/relay/ACK decisions, pubsub control plane, fanout codec, block exchange, and the shared-log receive pipeline — with JS keeping sockets, timers, signing, and the app-facing API. You still define and interact with programs in TS; there is no JS/Rust flip-flop on the message hot path.

Wire bytes, protocol ids, and observable protocol behavior are unchanged throughout. Every native path is an opt-in flag defaulting OFF; with flags off, the existing TS paths are byte-for-byte unchanged. Mixed js-only/rust-enabled fleets are exercised by dedicated mixed-topology specs at every layer.

## The arc

1. **`peerbit_wire` envelope codec + batched verification** (`nativeWire`): full DirectStream envelope decode/encode, byte-identical re-encode, sha256-prehashed Ed25519 batch verification, golden-vector parity suites in both directions. Same-tick inbound frames decode+verify in one native call (2.3–2.4× at realistic payloads).
2. **Receive fusion**: shared-log `RawExchangeHeadsMessage` payloads are recognized at the wire level and stashed in wasm linear memory (the codec core is embedded in the native-backbone module, so the one JS→wasm ingress write is shared); the prepared raw-receive pipeline consumes entry block bytes as in-wasm slices. No JS materialization of entries, no second boundary copy. Receive-prune bench: 72.6→61.0 ms (~16% end-to-end, the profiled `inputCopy` phase eliminated; 100-entry smoke 37.6→16.3 ms).
3. **DirectStream core in Rust** (`rustCore`): statement-level port of the state machine — multi-hop routes from ACK traces, session semantics, seen-cache dedup, relay/redundancy decisions, seek routing, and the 4-lane WRR outbound scheduler with byte-budget backpressure (byte buffers never cross the boundary; the core schedules `(sequence, byteLength, lane)` records). JS becomes a byte pump; the class contract, events, and constants are identical. Adversarially reviewed for behavioral parity: no blockers.
4. **DirectBlock port**: `/peerbit/direct-block/1.0.0` codec, provider resolution (negotiated→connected capped 32), hint cache and eager index in Rust; natively stored blocks are served via `block_response_payload` straight from the native store — served block bytes never materialize as a JS value.
5. **TopicControlPlane port**: all 8 `PubSubMessage` variants, root directory, shard mapping with the bit-exact `topicHash32` (including the f64 rounding quirk above 2^53), subscribe-state convergence rules. Mixed-topology spec covers subscription convergence, unsubscribe, and peer-unavailable eviction both directions.
6. **FanoutTree port**: the complete hand-written big-endian codec for all message kinds `MSG_JOIN_REQ(1)`…`MSG_PARENT_PROBE_REPLY(41)` at `/peerbit/fanout-tree/0.5.0` — encode byte-identical including JS numeric coercions and truncation-tolerance rules — plus the PR #911 parent-upgrade policy/gate decisions. Frame codecs flow through a single `FanoutWireCodec` seam (TS decoders extracted verbatim as the default). Adversarially reviewed including ~30k hostile differential fuzz cases and >6M decoder comparisons: no blockers; the two minor findings surfaced (falsy-zero `remoteSession` coercion, NaN-as-unset marshalling sentinel) are fixed on this branch with differential specs.
7. **Pure native path**: `Peerbit.create` gained a `network` option and `createRustPeerbitOptions()` enables the whole chain — wire-sync session installed as the pubsub inbound decoder, rust-core streams/pubsub/fanout/blocks, native-backbone data plane, shared-log native defaults advertised to programs.

## The thesis, proven mechanically

New E2E specs run two pure-native peers over real TCP syncing a TS-defined document store, with always-on wire counters asserting the hot path is JS-free. A 200-entry cold sync arrives as one wire batch and commits entirely inside wasm (stash → raw_receive → batch verify → graph/coordinates/document-index):

| counter | value |
|---|---|
| `sharedLog.rawReceive.jsEntryDecode` | **0** |
| `tsFrames` / `nativeFallbackFrames` | **0** / **0** |
| `tsSignatureVerifies` | **0** |
| stash `blockCopyOuts` / `bytesMaterialized` | **0** / **0** |
| `nativeFrames` (pubsub + fanout) | 257 + 13 |

The same spec runs a mixed leg (pure-native peer ↔ all-default peer, both directions) proving fleet interop; the document E2E additionally pins zero JS document decodes and zero generic JS index puts in strict native mode.

## What runs where (native preset)

| subsystem | Rust | JS |
|---|---|---|
| envelope decode+verify | batch codec + verification | header-only envelope object (routing state, app events); payload is a zero-copy view |
| dedup / routing / relay / ACK / seek | all decisions | timers, promises, sockets |
| outbound scheduling | 4-lane WRR + backpressure accounting | byte buffers, socket writes |
| pubsub | codec, subscribe rules, shard hashing, root directory | subscription maps, events |
| fanout | frame codec, parent-upgrade decisions | channel state machine, timers |
| block exchange | codec, provider rules, caches, native-store-served responses | store chain fallback |
| shared-log receive | wire stash → raw_receive → verify → index commit | high-level events only |
| transports | — | js-libp2p (noise/yamux/tcp/ws/webrtc) — native transports are a later phase |

## Enumerated exceptions (counted, not flip-flop)

1. One header-only TS envelope parse per frame (`tsEnvelopeDecodes`) feeding the TS routing state machine and app events; bytes enter wasm once and never come back.
2. The socket→wasm ingress copy per frame batch — inherent to a wasm engine; removable only by native transports.
3. The small fixed RPC shell decode per message; the per-entry inner payload resolves from the stash with zero JS decode.
4. One wasm→JS egress copy of the finished payload per outbound message (mirror of the receive ingress copy) and one TS ed25519 sign per outbound `DataMessage` (per ≤512 KB batch, not per entry). Both removable only by native transports.

Also found (pre-existing, reproduced at base): the prepared raw-receive path never dispatches program-level `onChange`, so raw sync defaults are gated to programs without a per-entry `onChange` consumer; document stores keep the plain exchange path (still native wire + native index commit). Real fix tracked separately.

## Measured end to end (5 runs ± stdev, TCP loopback, macOS arm64)

Cold bulk sync (8000 × 1KB entries, convergence from dial): **default 7.9s → native 3.3s (2.42×, 1007 → 2432 entries/s)**. The sender-side send loop is ~2% of wall on both legs, so the receive-fusion gain is fully visible and not masked by the unfused send path.

Sustained singleton live puts (2000 × 256B, awaited): **default 553 → native 2247 puts/s (4.06×)**, p50 visibility 2.0s → 0.6s. On the receive-only network plane (before send fusion was folded in) this leg *regressed* to 0.44× (279 vs 630 puts/s) — the folded-in send fusion (native outbound serialization + end-of-turn live-gossip coalescing behind a one-shot capability handshake) is what flips it, by amortizing the per-message wasm fixed costs across multi-entry raw frames while still flushing a lone put within one event-loop turn.

Stash pressure (128KB payloads, receiver dispatch gated to force the caps): above the 64MB byte cap, eviction fires and sync still converges on every run via the decode fallback plus synchronizer retries, at ~4.6× lower commit throughput across the eviction boundary. In normal ungated operation the caps are unreachable (dispatch backpressure holds stash depth at ~7–9 frames even for a 128MB burst). Harness: `benchmark:network-preset-e2e` in shared-log.

## Hardening round

A multi-agent adversarial review of the full diff (findings triple-verified before fixing) surfaced and fixed, with regression tests: wire-sync stash FIFO eviction dropping in-flight resolved sync messages (entries are now pinned between resolve and release); native signature verification switched to `verify_strict` (ed25519-dalek's `verify_batch`/`verify` accept small-order/weak-key signatures that libsodium — the TS path — rejects); a `usize` underflow permanently breaking the eager-block cache; the external-libp2p + `network` rejection now fires before storage/indexer/datastore acquisition (no leaked handles or locked directories); `LazyNativeLogBlockStore` now forwards `getBlockResponsePayload` (the native block-serving path was silently dead through the preset); shared-log native defaults advertise `nativeGraph: {optional: true}` instead of hard-failing program opens; undecodable block frames are dropped instead of thrown from the data listener; a throwing `resolveRequest` hook falls back to the decode path.

## Docs and browser coverage

As-built architecture (module topology, message flows, stash design, flag matrix, interop enforcement): `packages/transport/network-rust/ARCHITECTURE.md`. Browser CI legs in the native job: the wasm codec suites (68 tests) run under headless Chromium, and a two-browser-peer rust-core smoke over webrtc through a TS-core relay asserts `nativeFrames > 0` on both pages — the wasm engine verifiably runs the hot path in-browser.

## Tests

Every stage landed only when green: cargo 93+; network-rust JS 92 (default) and **427 passing under `PEERBIT_STREAM_RUST_CORE=1`** (stream+pubsub+blocks suites re-run against the native cores — the "both modes" legs, wired into CI); stream 164; pubsub 141; blocks 36; shared-log part-4 partition 115+ (incl. the new fused-path and E2E native-preset groups); native-backbone cargo + 59 JS; peerbit client 64 (incl. preset tests and external-libp2p rejection); browser legs 68 + rust-core smoke; packed-tarball smoke loads the full rust-core surface. Golden-vector suites pin the wire at every layer; two independent adversarial parity reviews (DirectStream, FanoutTree) found no blockers, and the hardening review's confirmed findings are all fixed on the branch.

---

## Consolidated scope (folded in)

This PR now also contains what were briefly split out as #990 and #991, so the native work is just **#939 → this PR**:

- **onChange dispatch from prepared raw receive** (was #990): document stores syncing under `rawExchangeHeads` now get index commits and change events (lazy entry views, materialization counted, zero-copy preserved for no-consumer programs). Also fixes a pre-existing `canAppend` ACL bypass in the native-validated join commit.
- **Send fusion** (was #991): outbound sync payloads serialize inside wasm from the native store, and live-put gossip coalesces into multi-entry raw frames with a one-shot capability handshake. Live puts **4.06× over default** (was a 0.44× regression on the receive-only branch), cold sync 2.58×.

Folding them together also closes the one interaction the send-fusion review flagged: the live-append path now dispatches onChange through the same unified receive→join path, so the gap is closed rather than extended. Both features coexist as independent code paths — verified against both parent branches, with all 6 new specs (2 onChange + 4 send-fusion) green together in part-4 and the document E2E green in part-2a.


peerbit-org:
# PR #988 reviewer guide — native network plane

- **PR**: dao-xyz/peerbit #988, branch `feat/network-rust-wire`
- **Base**: `feat/native-log-append-kernel` (PR #939 — native codec/batching/sync-planner/raw-receive kernel)
- **Size**: 34 commits, 84 files, +20 672 / −505 (plus the follow-up docs commit `83b32a49b` added after the fact)
- **As-built architecture**: `packages/transport/network-rust/ARCHITECTURE.md` (committed on the branch as `83b32a49b` — read it first; this guide maps the commits onto it)

Get the diff:

```
git -C <worktree> fetch origin feat/native-log-append-kernel feat/network-rust-wire
git -C <worktree> log --oneline origin/feat/native-log-append-kernel..origin/feat/network-rust-wire
git -C <worktree> diff --stat origin/feat/native-log-append-kernel..origin/feat/network-rust-wire
gh pr view 988
```

Global invariants every commit was written under (check them, they are the review contract):

1. Wire bytes, protocol ids and observable protocol behavior are identical in both modes.
2. Every native path is opt-in and defaults OFF; with flags off the TS paths are byte-for-byte unchanged (the only unconditional additions are the plain-integer `wireCounters` increments and the `sharedLog.rawReceive.jsEntryDecode` profile event).
3. Native packages must be registered in BOTH `pnpm-workspace.yaml` and the root `package.json` `workspaces` list (commit 23 below is the fix for exactly this).
4. Benchmarks are never run concurrently with builds/tests.

---

## Commit map by arc stage (oldest first)

### Stage 1 — wire codec bootstrap (`nativeWire`)

| # | commit | subject | review notes |
|---|---|---|---|
| 1 | `24b6aad7e` | feat(stream): optional native wire batch decode and verify | The `DirectStreamOptions.nativeWire` seam: same-tick frame collection in `processRpc`, memoized `_verified` seeding, per-frame TS fallback. **Attention**: the fallback conditions (decode failure, secp256k1, non-sha256 prehash) and that the flag-off path is untouched. |
| 2 | `8208aab9b` | feat(network-rust): peerbit_wire envelope codec and batched verification | New workspace package + crate: hand-rolled borsh codec (incl. the retired SeekDelivery gap at variant 2), signable-bytes rule, batched Ed25519. Golden-vector parity both directions. **Attention**: `wire.rs` offset arithmetic vs the TS hand-rolled readers; the signable range (mode + signature option flags zeroed). Superseded in part by commit 25 (strict verify). |
| 3 | `1ef804fd2` | ci: cover @peerbit/network-rust in the native test job | Mechanical: CI + release-please + packed-tarball smoke. |

### Stage 2 — receive fusion (`wireSync`)

| # | commit | subject | review notes |
|---|---|---|---|
| 4 | `0e33cf4ce` | feat(native-backbone): wire-sync session for fused raw exchange-head receive | **Highest-risk file of the PR**: `packages/utils/native-backbone/src/wire_sync.rs`. The sync-payload recognizer (exact borsh nesting, plaintext-only, `respondTo` absent), the local-delivery check mirroring `TopicControlPlane.onDataMessage`, the FIFO bounds (512 msgs / 64 MiB), and the `prepare_stashed_raw_receive_*` twins. Read together with commit 24 (pinning) — the lifecycle is not complete without it. |
| 5 | `9de91624f` | feat(shared-log): fuse wire-stashed raw exchange heads into native receive | The RPC `resolveRequest` hook, `StashBackedRawExchangeHeadsMessage` with lazy byte materialization, stash release at the end of `onMessage`, and the fallback ladder (session/backbone/stash entry/native method missing → borsh path). **Attention**: every exit path of `onMessage` must release; lazy `bytes` getters must not be triggered by validation (they probe `byteLength`). Read together with commit 28 (throwing hook fallback). |
| 6 | `ca19055c9` | bench(shared-log): wire-path raw receive legs with and without fusion | Mechanical (bench only). Numbers: 72.6→61.0 ms per 1000-entry batch; `inputCopy` phase eliminated. |

### Stage 3 — DirectStream core (`rustCore`)

| # | commit | subject | review notes |
|---|---|---|---|
| 7 | `088fd531e` | feat(network-rust): port the direct-stream state machine core | Statement-level port: `routes.rs` (largest single file, 1076 lines), `seen_cache.rs`, `lanes.rs`, `decisions.rs`. Mostly mechanical transcription — verify via the suite re-run (below) rather than line-by-line; spot-check the externalized time/timer seams (`now_ms` params, `cleanup_requested`). |
| 8 | `f99fd0245` | feat(stream): rust-core mode delegating protocol decisions to the native core | **Attention**: this touches the live TS hot path. Every decision call-site branches on `this.rustCore`; the flag-off branch must be the pre-existing code verbatim. Check option precedence: `nativeWire ?? rustCore.nativeWire`; `rustCore: false` opts out of the test-time injection; the batched inbound path resolves `processRpc` only after the frame's `processMessage` turn. |
| 9 | `7308098c4` | feat(network-rust): rust-core adapter, mixed topology and suite re-run | The `createRustCoreStream()` adapter (RoutesLike timer scheduling, sequence-mapped lanes so bytes stay in JS) + the mixed-topology spec + the `PEERBIT_STREAM_RUST_CORE=1` re-run harness. **Attention**: the RoutesLike adapter's snapshot materialization and single-timer scheduling parity. |

### Stage 4 — protocol ports (blocks → pubsub → fanout, by increasing risk)

| # | commit | subject | review notes |
|---|---|---|---|
| 10 | `aeb029f64` | feat(network-rust): port the block-exchange protocol core | Codec + provider rules + FIFO-cache port with the `@peerbit/cache` lazy-delete quirk. Mostly mechanical; the cache-quirk port produced the bug fixed in commit 26 — review `FifoCache` accounting there. |
| 11 | `7ad4b8aaf` | feat(log-rust): serve block responses from the native store | `block_response_payload`: BlockResponse serialized inside wasm. Small; check the crate dependency direction (log-rust → peerbit_wire, codec owned by one crate). |
| 12 | `2badf7888` | feat(blocks): rust-core mode for DirectBlock and RemoteBlocks | Seam commit; `DirectStream.rustCore` becomes protected. **Attention**: the optional `getBlockResponsePayload` resolution through the store chain (see commit 31 for the preset-level miss). |
| 13 | `cbb85bf60` | feat(network-rust): port the topic-control protocol core | Codec + `topicHash32` (bit-exact f64 rounding above 2^53) + root directory + convergence rules. Transcription verified by fuzz/golden specs. |
| 14 | `81dd366ab` | feat(pubsub): rust-core mode for the TopicControlPlane | Seam commit. **Attention**: the subscription maps stay authoritative host-side (public API) — the native core must only feed decisions, never own that state. |
| 15 | `a5d4aaea8` | feat(network-rust): port the fanout-tree protocol core | The largest codec surface (MSG 1–41, big-endian, JS numeric coercions, skip/truncate tolerance) + PR #911 parent-upgrade decisions. Mechanical *because* of the fuzz/golden coverage; do not line-review 41 message kinds — review the test strategy in `test/fanout-codec.spec.ts` instead. |
| 16 | `f4388db36` | feat(pubsub): rust-core mode for the FanoutTree | **Attention**: the `FanoutWireCodec` seam extraction — the TS decoders were moved verbatim from inline `onDataMessage` parsers into `fanout-tree-codec.ts` and become the default implementation. Verify the extraction is behavior-preserving (this is the one place "flag-off unchanged" means *moved* code, not untouched code); multiaddr bytes stay raw across the boundary so `multiaddr()` validation runs identically in both modes. |

### Stage 5 — the pure native path

| # | commit | subject | review notes |
|---|---|---|---|
| 17 | `bde444493` | feat(stream): wire counters and out-of-band native envelope verification | `wireCounters` (always-on, plain increments — verify no behavior rides on them) + `seedNativeWireVerification` for fanout-delivered shard frames; TS verification stays authoritative on native failure. |
| 18 | `72d5393b5` | feat(shared-log): client-advertised native defaults and JS entry-decode counter | **Attention**: the defaults-merging rule (explicit per-open options, including `false`, always win; absent the client property nothing changes) and the onChange gate (raw sync defaults only for programs without a program-level `onChange` consumer — the prepared raw receive never dispatches it). |
| 19 | `7977e5f29` | feat(peerbit): native network preset wiring the whole chain end to end | `Peerbit.create` `network` option + `createRustPeerbitOptions` composition + the counter E2E specs (shared-log and document). **Attention**: identity must exist before services (wire-sync keyed by public key hash); one rustCore shared across the three streams; wireSync installed only on pubsub. |

### Stage 6 — review findings and hardening fixes

| # | commit | subject | origin |
|---|---|---|---|
| 20 | `1337a7956` | fix(network-rust): coerce a stored remote session of 0 to -1 in routes add | Adversarial review finding (see below) |
| 21 | `e8c4b4bba` | fix(network-rust): distinguish set from unset parent-upgrade options | Adversarial review finding (see below) |
| 22 | `0f38090ca` | fix(pubsub): assert subscription presence on the keep-existing path | CI (ts-node strict type-check at module load) |
| 23 | `9c1acda1f` | fix(ci): register @peerbit/network-rust in the npm workspaces list | CI (root build graph; global invariant 3) |
| 24 | `b5a6a08c5` | fix(native-backbone): pin resolved wire-sync stash entries against eviction | **Correctness-critical**: without pinning, FIFO eviction during the async gaps of `onMessage` silently drops the heads of the in-flight message (no TS fallback after resolve). Cargo tests: pinned entry survives a flood past the 512 cap; duplicate-of-pinned keeps the pin. |
| 25 | `b06e2ea0a` | fix(network-rust): verify direct-stream signatures with strict checks | **Correctness-critical**: `verify_batch`/non-strict verify accept weak-key/small-order signatures libsodium rejects; a native accept is memoized and suppresses the TS re-check, so accept-direction divergence had no fallback. Now per-signature `verify_strict`. |
| 26 | `cf7eb7178` | fix(network-rust): guard the eager-block FIFO cache against usize underflow | del/add/del cycle on one cid → wrap to `u32::MAX` on wasm32 release → cache evicts everything forever. `saturating_sub`, matching the TS cache's benign miscount. |
| 27 | `de0b0e2b2` | fix(network-rust): coerce a stored remote session of 0 to -1 in update_session | Same falsy-coercion class as #20, applied to `update_session`. |
| 28 | `a8cbd9414` | fix(rpc): fall back to the decode path when the resolveRequest hook throws | A throwing user `sync.profile` callback inside the hook dropped raw exchange-heads messages the unfused path would have processed. |
| 29 | `9b81076ff` | fix(blocks): drop undecodable data frames instead of throwing from the listener | rust-core strict UTF-8 reader threw out of an EventTarget listener (uncaughtException) where pure TS was lossy-tolerant. |
| 30 | `63cbd71c6` | fix(peerbit): validate network+external-libp2p early and soften native graph defaults | Resource-leak-on-throw (LEVEL_LOCKED on retry) + `nativeGraph: { optional: true }` so a missing optional native module degrades instead of aborting every program open. |
| 31 | `72d6037c3` | fix(peerbit): forward native block serving and warn on non-persistent log blocks | `LazyNativeLogBlockStore` did not forward `getBlockResponsePayload`, so the native serve path never engaged through the preset; plus a data-loss warning for `nativeLogBlocks` + `directory`. |

### Stage 7 — macro benchmark

| # | commit | subject | notes |
|---|---|---|---|
| 32 | `6340c055f` | bench(shared-log): end-to-end network preset benchmark | `benchmark/network-preset-e2e.ts` + README. Honest-negative documented: singleton live-puts are ~0.4× on the native preset (send fusion is the identified fix); cold-sync ~2.4×; stash-pressure ~4–5× commit cost above the eviction boundary while still converging. |

---

## Where review risk concentrates

Spend your time here (roughly in order):

1. **Stash lifecycle** — `wire_sync.rs` (commits 4, 24) + the shared-log consume side (5, 28). The invariant to convince yourself of: *eviction is loss-free for never-resolved entries, and resolved entries can never be evicted before release*. Trace: stash → resolveRequest → stashedMeta (pin) → onMessage awaits → release; and the miss path → borsh decode.
2. **Verification semantics** — commits 2 and 25. The accept set of the native verifier must equal libsodium's exactly; the memoized `_verified` makes any native false-accept unrecoverable. Check the `verify_strict` reasoning and the weak-key cargo test.
3. **The rust-core seams in live TS hot paths** — commits 8, 14, 16 (and 12). The pattern is always "branch to the core when enabled, pre-existing code otherwise"; the risk is a seam that changed flag-off behavior. Commit 16 is special: TS decoders were *moved* (extracted verbatim) — diff the extraction against the removed inline parsers.
4. **Preset composition and defaults merging** — commits 18, 19, 30, 31. Option precedence (`nativeWire ?? rustCore.nativeWire`, explicit per-open options win, `sharedLogDefaults !== false`), the onChange gate, resource acquisition order in `Peerbit.create`.
5. **Falsy/NaN marshalling at the boundary** — commits 20, 21, 27. The class of bug: TS reads state through `|| -1` / `??` and the port compared raw values. Worth asking: are there other `||`-coerced reads in `routes.ts`/`fanout-tree.ts` the ports did not mirror? (The suite re-runs and differential specs are the systematic answer.)

Mostly mechanical transcription — verify via the test strategy, skim the code:

- `routes.rs`, `seen_cache.rs`, `lanes.rs`, `decisions.rs` (commit 7): covered by the 427-test both-modes re-run + differential routes specs.
- `fanout_tree.rs` (15), `topic_control.rs` (13), `block_exchange.rs` (10): golden byte-identity vectors, decoder parity on every prefix of every frame and on random garbage, hash-parity fuzz corpus, ~30k hostile differential fuzz cases on fanout.
- CI/workspace plumbing (3, 22, 23) and the two bench commits (6, 32).

---

## Reproducing every claim

Commands marked ✔ were re-run during the writing of this guide on the
branch head and passed. All paths are relative to the repo root; build
prerequisites: `pnpm install`, plus a Rust toolchain with `wasm-pack` for
the native packages. Never run the benchmarks concurrently with builds or
tests.

### Cargo (host) tests — the JsValue-free cores

```
(cd packages/transport/network-rust && cargo test)     # ✔ 93 passed
(cd packages/utils/native-backbone && cargo test)      # ✔ passes (incl. wire_sync stash/pinning tests)
```

### Golden vectors, mixed topologies, and the both-modes re-runs

```
pnpm --filter @peerbit/network-rust run test
#   = cargo test + wasm build + the default-mode aegir suites:
#     parity.spec.ts ("wire parity", both directions), topic-control.spec.ts,
#     fanout-codec.spec.ts (incl. "parent-upgrade parity (PR #911)"),
#     block-exchange.spec.ts, stream-integration.spec.ts, and the
#     mixed-topology specs (directstream / pubsub / fanout rust-core mixed topology)

pnpm --filter @peerbit/network-rust run test:stream-rust-core
#   = PEERBIT_STREAM_RUST_CORE=1: re-runs the unmodified @peerbit/stream,
#     @peerbit/pubsub and @peerbit/blocks behavioral suites against the
#     injected native core (427 passing at merge time). Wired into CI
#     (.github/workflows/ci.yml, "Test @peerbit/network-rust" steps).
```

### Default-mode suites (flag-off paths unchanged)

```
pnpm --filter @peerbit/stream run test        # 164
pnpm --filter @peerbit/pubsub run test        # 141
pnpm --filter @peerbit/blocks run test        # 36
pnpm --filter @peerbit/rpc run test           # incl. the throwing-resolveRequest fallback spec
pnpm --filter @peerbit/native-backbone run test
pnpm --filter peerbit run test:node           # client, incl. preset tests + external-libp2p rejection
```

### The counter E2E specs (the "thesis, proven mechanically")

```
# shared-log: two pure-native peers over real TCP, 200-entry cold sync,
# asserts jsEntryDecode=0, tsFrames=0, nativeFallbackFrames=0,
# tsSignatureVerifies=0, stash blockCopyOuts=0, full stash release; plus a
# mixed pure-native <-> all-default leg in both directions.
cd packages/programs/data/shared-log
pnpm test -- --no-build --grep "network e2e native preset"     # ✔ 2 passing
# (drop --no-build, or build the workspace first, on a fresh checkout)

# document store in strict native mode (adds zero JS document decodes and
# zero generic JS index puts), pure-native pair + mixed pair:
cd packages/programs/data/document/document
pnpm test -- --no-build --grep "native network e2e"

# raw exchange-head sync unit/integration specs (fused path incl. the
# never-stashed fallback):
cd packages/programs/data/shared-log
pnpm test -- --no-build --grep "raw exchange-head sync"
```

The CI partition that carries the shared-log groups (115 tests incl. the
`network` and `raw exchange-head sync` groups):

```
pnpm run test:ci:part-4
```

### Packed-tarball smoke (shipped-artifact surface)

```
node ./scripts/native-pack-smoke.mjs
```

### Benchmarks (sequential only, quiet machine)

```
# micro: TS vs native batch decode+verify on the same corpus (2.3-2.4x claim)
pnpm --filter @peerbit/network-rust run benchmark

# receive path fused vs unfused wire legs (72.6 -> 61.0 ms claim)
cd packages/programs/data/shared-log
pnpm run benchmark:receive-prune

# end-to-end preset benchmark (cold-sync ~2.4x, live-puts ~0.4x,
# stash-pressure claims); see benchmark/README.md for knobs
pnpm run benchmark:network-preset-e2e
NET_BENCH_SCENARIOS=cold-sync NET_BENCH_LEGS=native pnpm run benchmark:network-preset-e2e
```

---

## The two adversarial parity reviews

Both were independent, dedicated reviews of the ported state machines
against the TS originals, run before the fix stage landed.

**Review 1 — DirectStream state-machine port** (commits 7–9: routes,
session semantics, seen-cache, relay/redundancy, seek routing, WRR lanes).
Outcome: **no blockers, no findings requiring code change**. The behavioral
surface is additionally pinned by the full `@peerbit/stream` suite re-run in
rust-core mode.

**Review 2 — FanoutTree port** (commits 15–16), including ~30 000 hostile
differential fuzz cases and >6 000 000 decoder comparisons across both
implementations. Outcome: **no blockers; two minor findings**, both fixed on
this branch with differential specs:

1. *Falsy-zero `remoteSession` coercion.* TS reads the stored remote session
   through `prev.remoteSession || -1`, so a stored value of exactly `0` is
   coerced to `-1` for restart detection and `Math.max` merging; the port
   compared raw stored values, so a mixed fleet could disagree on restart
   detection. Fixed verbatim in `Routes::add` (`1337a7956`, with a cargo
   test and RoutesLike-contract mirror specs), and the same class was then
   swept and fixed in `update_session` (`de0b0e2b2`).
2. *NaN-as-unset marshalling sentinel.* The parent-upgrade options crossed
   the boundary with NaN meaning "absent", making an explicitly-NaN option
   indistinguishable from an unset one — TS keeps an explicit NaN (silently
   disabling `> 0` gates) while the native core normalized it to defaults.
   Fixed in the marshalling (`e8c4b4bba`): a presence bitmask marks provided
   options and values travel verbatim; `floor_max` mirrors JS `Math.max` NaN
   poisoning. Differential specs pin the three repro fields, the NaN
   cascade, and identical gate outcomes in both cores.

Note the review asymmetry when weighing the fixes: the finding class was
*boundary marshalling / JS coercion mirroring*, not protocol logic — the
codec byte-identity and decision parity survived the fuzz volume intact.

---

## Quick sanity checklist before approving

- [ ] `git diff origin/feat/native-log-append-kernel..HEAD -- packages/utils/indexer/rust/src/index.ts packages/utils/indexer/tests` is empty (files are off-limits).
- [ ] `@peerbit/network-rust` present in BOTH `pnpm-workspace.yaml` and root `package.json` `workspaces`.
- [ ] With all flags off: only additive/observability lines touch the hot paths (`wireCounters` increments, `jsEntryDecode` profile event).
- [ ] Protocol id strings grep to the same three values as on base.
- [ ] CI native job runs cargo tests, the parity suite, the rust-core re-run, and the pack smoke (`.github/workflows/ci.yml` around line 496).

PR #990, opened 2026-07-03
fix(shared-log): dispatch program onChange from the prepared raw receive

peerbit-org:
Fixes the raw-receive onChange gap (stacked on #988) and removes the workaround that routed document stores around the raw sync path.

## The bug

The prepared raw-receive path committed entries via a hash-only sink and never dispatched program-level `onChange` events — a document store syncing under `rawExchangeHeads` silently missed index commits (log filled, index stayed empty). Pre-existing at the #939 base; #988 mitigated it by gating raw sync defaults to programs without a per-entry onChange consumer.

## The fix

Change-event dispatch from prepared joins (design (b); native document projection inside wasm remains the follow-up recorded in ARCHITECTURE.md §10 — it only covers strict-native document stores, so it cannot replace the generic contract restored here):

- When a program registers an onChange consumer, `SharedLog.onMessage` no longer passes the hash-only sink, so the lower log dispatches the program-level change event. Dispatched entries are **lazy views** (`PreparedRawExchangeEntry`): meta reads come from prepared raw facts; block bytes and the borsh decode materialize only when the consumer touches payload/signatures/storage bytes (`Entry.prepareMultihashBytesLazy` keeps even prepared-block registration lazy over the wire stash). Materialization is counted (`jsEntryDecode` with `details.lazy`, stash `blockCopyOuts`/`bytesMaterialized`) — the fused no-consumer path stays assertable at zero.
- **Found and fixed while removing the gating**: the native-validated join commit bypassed program-level `canAppend` hooks (pre-existing at the #939 base, masked for document stores only by the gating). Programs with a hook now take the lower-log batch join where the hook fires per entry, with signatures still verified natively during prepare — without this, removing the gating would have silently bypassed document `canPerform` ACLs.
- The onChange-consumer gating is removed from the client-advertised native defaults; document stores inherit `rawExchangeHeads`+`nativeWireSync` from the preset.

## Verification

Bug-shape repro spec verified both directions (dispatch disabled: log fills, index stays 0; with fix: index commits land, change sets correct, raw wire path asserted). Full gate green: workspace build; shared-log part-4 mock run 117 passing (zero-decode invariants intact); document part-2a suites 344 + 3 passing — including the rewritten native E2E exercised end to end (program topic registers on the wire-sync session, consumer-driven lazy materialization is exactly what surfaces as stash copy-outs, index still commits natively); native-backbone cargo 11 + JS 59; log 231 (covers the lazy prepared-block API); peerbit client 66 (covers the default changes); eslint clean on the diff. No spec or source adjustments were needed during final verification.

### DIFF
--- a/packages/log/src/entry.ts
+++ b/packages/log/src/entry.ts
@@ -89,6 +89,29 @@ const preparedEntryBlockFromBytes = (
 	};
 };
 
+const preparedEntryBlockFromBytesSource = (
+	bytesSource: () => Uint8Array,
+	cid: string,
+): PreparedEntryBlock => {
+	let cidObject: ReturnType<typeof cidifyString> | undefined;
+	let bytesValue: Uint8Array | undefined;
+	return {
+		block: {
+			get bytes() {
+				return (bytesValue ??= bytesSource());
+			},
+			get cid() {
+				cidObject ??= cidifyString(cid);
+				return cidObject;
+			},
+			get value() {
+				return (bytesValue ??= bytesSource());
+			},
+		} as PreparedEntryBlock["block"],
+		cid,
+	};
+};
+
 interface Meta {
 	clock: Clock;
 	gid: string; // graph id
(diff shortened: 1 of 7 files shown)


peerbit-org:
Folded into #988 (the consolidated pure-native-path PR) as commits d57436d76, 0d6936a93, f5df135eb — cherry-picked on top of the network plane and re-verified with send fusion. Closing; the change ships in #988.

PR #991, opened 2026-07-03
Native send fusion: outbound serialization from the store and live-put coalescing

peerbit-org:
Completes the outbound half of the native network plane (stacked on #988). With this, the message hot path is JS-free in **both** directions: receive was fused in #988; send now serializes inside wasm straight from the native block store, and live-put gossip coalesces into multi-entry raw frames that take the receiver's fused stash path.

## What's here

- **Native outbound serialization**: `peerbit_wire::sync_payload` gained a single-pass, exactly-sized, borrowed-heads encoder (golden-pinned byte-identical to the TS `serialize(PubSubData(RequestV0(DecryptedThing(RawExchangeHeadsMessage))))` layout); native-backbone's `sync_send.rs` builds the complete payload in wasm from the native block store — entry block bytes never surface as JS values on the send path. Pubsub gained `publishPreEncodedData()` so the finished payload becomes a signed `DataMessage` without the re-wrap copy. One TS ed25519 sign per ≤512KB message remains (per-batch, not per-entry; the node identity key stays out of wasm).
- **Raw-capability negotiation + live-put coalescing**: peers with `sync.rawExchangeHeads` advertise a one-shot `SyncCapabilitiesMessage([0,10])` in the existing subscribe flow (opt-in-only variant; unknown-variant receivers drop it via the existing namespace guard — default wire behavior byte-for-byte unchanged). Silent fire-and-forget append gossip to all-capable recipient sets coalesces per recipient set until end of the current event-loop turn (caps: 256 entries / 128KiB, under the 512KiB raw cap) — a lone put flushes within one turn; awaited-put bursts ship as multi-entry raw frames. Non-advertising peers keep the byte-identical plain path. Bulk sync responses and repair pushes route through the same fused seam.
- **The "3× slower native append" from the #988 bench was a measurement artifact**: isolated, the native singleton append is ~1.2× *faster* than default (0.139 vs 0.158–0.183 ms); the old figure was receiver-side per-message work interleaving into the shared benchmark thread — which the coalescing removes.
- E2E counter specs extended: cold sync asserts fused-send entries ≥ N and zero JS outbound block-byte copies; a new live-puts leg asserts zero plain live sends, coalescing (messages < entries), and fused receive of live frames (`jsEntryDecode=0`, `blockCopyOuts=0`, `tsFrames=0`). Irreducible send-path exceptions documented in ARCHITECTURE.md §8: one wasm→JS egress copy of the finished payload per message (mirror of the receive ingress copy) and one TS sign per message.

## Measured (`benchmark:network-preset-e2e`, 2 warmups + 5 measured runs, mean ± stdev, TCP loopback, arm64 mac)

| leg | default | native | ratio |
| --- | --- | --- | --- |
| live puts (2000 × 256B awaited), puts/s | 553 ± 67 | 2247 ± 958 | **4.06×** |
| live puts p50 visibility | 2024 ms | 623 ms | 3.2× lower |
| cold sync (8000 × 1KB), entries/s | 923 ± 32 | 2385 ± 564 | **2.58×** |

The #988 baseline for live puts was **0.44×** (279 vs 630 puts/s) — this PR flips the native preset's one regression into its largest win. Native live-put variance is high (~43% CV); the ratio holds with wide margin across all runs. Stash-pressure legs (128KB payloads): below-cap 101 MB/s commit; forced above the 64MB FIFO cap (2434 evictions/run, stash pinned at 127 frames) throughput degrades ~6.3× but converges on every run with zero deserialize fallbacks — evicted frames recover through the synchronizer retry path.

## Verification

shared-log part-4 mock run 120 passing (incl. 6 new send-fusion specs); native-backbone cargo 13 + JS 59; network-rust cargo 94 + JS 92 default + 428 under `PEERBIT_STREAM_RUST_CORE=1`; log 231; peerbit client 66; rpc 26; pubsub 141; cargo fmt + root eslint clean on the diff. Adversarial review: no blockers or majors; two minors noted for follow-up — the capability handshake extends the (pre-existing) onChange gap to live appends for explicitly-configured programs (dissolves when the onChange-dispatch PR merges), and repair-hinted raw frames deserve dedicated coverage.

### DIFF
--- a/packages/programs/data/shared-log/benchmark/README.md
+++ b/packages/programs/data/shared-log/benchmark/README.md
@@ -39,15 +39,17 @@ Defaults: 2 warmup + 5 measured runs per leg, strictly sequential. Do not
 run anything else (builds, tests, other benchmarks) concurrently.
 
 Indicative shape of the results (one dev machine, defaults; absolute
-numbers vary): cold-sync ~2.4x entries/s for the native preset with the
-sender-side send loop at ~2% of wall on both legs (outbound send does not
-hide the receive win). Sustained singleton live-puts currently run slower
-on the native preset (~0.4x throughput, higher visibility lag) — the
-per-message fixed cost of the fused receive dominates for one-entry
-messages, which is what outbound send fusion / batching is expected to
-address. Above the stash cap the commit phase pays ~4-5x in MB/s versus
-just below it (eviction fallback + the synchronizer's retry duplicates)
-while still converging on every run.
+numbers vary): cold-sync well above 2x entries/s for the native preset
+with the sender-side send loop at ~2-3% of wall on both legs. Sustained
+singleton live-puts run several times faster on the native preset with the
+fused send: awaited puts coalesce into multi-entry raw frames serialized
+inside wasm, which amortizes the receiver's per-message fixed costs that
+previously made this leg ~0.4-0.9x of the default (one instrumented run on
+one dev machine: ~6x default throughput with ~7x lower p50 visibility
+latency; before send fusion the same machine measured ~0.9x). Above the
+stash cap the commit phase pays ~4-5x in MB/s versus just below it
+(eviction fallback + the synchronizer's retry duplicates) while still
+converging on every run.
 
 ```
 cd packages/programs/data/shared-log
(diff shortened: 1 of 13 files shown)


peerbit-org:
Folded into #988 (the consolidated pure-native-path PR) as commits e61bbcfcf, 24b85f0a6, babcd7ec6, 108c5955d, 6dcb4a978. Folding it together with the onChange dispatch (was #990) also closes the live-append onChange interaction this PR's review flagged. Closing; the change ships in #988.

PR #994, opened 2026-07-05
[spike] rust-libp2p node interop feasibility (not for merge)

peerbit-org:
**Feasibility spike — not for merge.** A throwaway, workspace-isolated proof (`spikes/rust-libp2p-poc/`, not a cargo/pnpm workspace member, so CI never touches it) answering: can a rust-libp2p node swarm replace js-libp2p on node peers while still interoperating with the existing js-libp2p fleet — and what would it cost?

## Verdict: FEASIBLE-WITH-CAVEATS (node only)

The shared interop path (TCP + WebSocket + Noise + Yamux + multistream-select 1.0 + identify + circuit-relay v2) is in libp2p test-plans' continuously cross-tested known-good set (rust 0.56 ↔ js 3.x), and **the spike ran it live**:

- **Goal 1** — two rust-libp2p peers connect over TCP, negotiate Noise+Yamux+identify, open `/peerbit/direct-stream/2.0.0` via `libp2p-stream`, and round-trip a varint-length-prefixed `DataMessage` frame. ✅
- **Goal 2 (the real question)** — a rust-libp2p 0.56 peer dials a **live js-libp2p 3.3.4 node** (matching Peerbit's exact versions: noise 17, yamux 8, tcp 11, identify 4.1), negotiates TCP+Noise+Yamux+multistream cross-implementation, opens `/peerbit/direct-stream/2.0.0` on the js node, sends a varint-framed `DataMessage` that **js decodes byte-identically**, and receives the echo back. ✅

Full report with the per-component interop matrix, the concrete interop contract from Peerbit's live config, the engine-binding design, and a phased plan: [`spikes/rust-libp2p-poc/FEASIBILITY.md`](spikes/rust-libp2p-poc/FEASIBILITY.md).

## The load-bearing caveats

1. **The transport swap is the easy half.** The real work is speaking Peerbit's DirectStream wire (Borsh schema, Ed25519/SHA-256 signing, seen-cache, 4-lane scheduler, routing) natively over the **three** frozen multicodecs (`/peerbit/direct-block/1.0.0`, `/peerbit/topic-control-plane/2.0.0`, `/peerbit/fanout-tree/0.5.0`) — much of which #988 already ported; what's missing is binding a rust-libp2p swarm to that engine instead of JS being the byte pump.
2. **`libp2p-stream` is 0.4.0-alpha** — API-churn risk; documented fallback is a hand-rolled `NetworkBehaviour`.
[...]

PR #995, opened 2026-07-05
[draft] @peerbit/transport-rust: native rust-libp2p transport + profiling + fully-native-node feasibility

peerbit-org:
**Phase 1 of the long-term node rust-libp2p transport — draft, opt-in, additive.** Purely additive (5124 insertions, 0 deletions): the default js-libp2p node path is untouched and this package is inert unless explicitly used. Follows the spike (#994), which proved rust↔js interop is feasible.

## What this is

`@peerbit/transport-rust` (crate `peerbit_transport`) — the node transport that owns the rust-libp2p connection layer natively, so node peers can eventually stop pumping bytes through JS. It reuses the frozen `peerbit_wire` codec (path dep, not re-implemented) and speaks the exact `/peerbit/*` wire, so a rust-libp2p node interoperates with the existing js-libp2p fleet.

## What Phase 1 builds

- rust-libp2p 0.56 connection layer matching Peerbit's **node** config: TCP + WebSocket(+DNS) + Noise + Yamux + identify + circuit-relay-v2 client; yamux-only muxer, noise-only encrypter.
- **One Ed25519 key drives both** the libp2p peerId (byte-identical to Peerbit's derivation) and the wire signing key.
- The **three frozen multicodecs** — `/peerbit/direct-block/1.0.0`, `/peerbit/topic-control-plane/2.0.0`, `/peerbit/fanout-tree/0.5.0` — mounted over `libp2p-stream`, framed with it-length-prefixed unsigned-varint calling `peerbit_wire` directly on socket slices.

## Interop: proven live

A rust `peerbit_transport` peer dialed a **live js-libp2p 3.3.4 Peerbit-config node**, negotiated TCP+Noise+Yamux+identify cross-implementation, opened all three `/peerbit/*` streams, and per stream sent a **signed `DataMessage` envelope from the real `peerbit_wire` codec** that the js side decoded with production `@peerbit/stream-interface` `Message.from` and re-serialized byte-identically (232/239/231-byte envelopes); rust re-decoded and **signature-verified** the echo. Byte parity both directions, all three streams. (Adversarial review reproduced this independently.)

## Non-breaking / CI-safe

The crate is a native `rlib`+`bin`, **never** wasm-packed, not in any default cargo/wasm build. `network-rust` cargo test still 94/0; the existing workspace build and CI partitions are unchanged (the transport CI partition discovers the package but runs zero Rust — a verified no-op); only the native job gains an opt-in cargo-test + live-interop step.

## The decision this PR surfaces (gates Phase 2)

The swarm needs **native tokio sockets**, which wasm-on-node cannot own — so binding it into the wasm-based node runtime requires choosing: **(1)** relax the no-napi rule (in-process napi addon — the *only* option that removes the socket→wasm copy the whole track exists to remove); **(2)** a sidecar process over IPC (honors no-napi, keeps the swarm native, but bytes still cross a boundary so no zero-copy); or **(3)** defer the binding, ship this crate + interop, and revisit after Phase-0 profiling (which the plan and FEASIBILITY.md both require before committing). Phase 1 is built native-only and is valuable under all three.

## Deferred to Phase 2+

TS adapter + node client wiring, the wasm-engine binding (per the decision above), relay/NAT hardening, WebSocket live interop (compile-checked; TCP is the proven path), and ultimately replacing js-libp2p on node. Browser stays js-libp2p permanently (no browser↔browser `/webrtc` upstream) — the target is a hybrid fleet.

Review found no blockers; 3 minors (harness doesn't exercise the js verify path, a `Message.from` vs `DataMessage.from` doc nit, and `peerbit_wire`'s wasm-bindgen deps leak into the native build uncfg-gated — compiles today, worth hardening).

---

## Also in this PR: fully-native node feasibility (folded from the spike)

To keep the native-transport exploration in one place, this PR also carries the **fully-native Peerbit node** feasibility proof (`spikes/native-node/`, isolated crate — see [`FEASIBILITY.md`](spikes/native-node/FEASIBILITY.md)).

**Verdict: FEASIBLE for the transport + network-engine layer — proven by running it.** Two native `peerbit_transport` nodes in one process push a real signed `PubSubData` through the full native receive engine (decode+verify → dedup → decision → `topic_control` decode → ack → lane → signed ack, verified natively) with **no JS, no wasm, no napi, no sidecar** — `nm` shows 0 `__wbindgen` symbols (cores link native, wasm glue stripped). Because the engine cores are JsValue-free by design, a fully-native node has no wasm boundary at all, so the napi-vs-sidecar question dissolves. Rust programs then sit on top as a thin trait over the native log+index, coexisting with TS programs over the same frozen wire.

Remaining known-shape work (in `FEASIBILITY.md`): re-author `native-backbone`'s data-plane orchestration natively, a native persistence store, and mixed-fleet parity. The hard/uncertain part (native engine + no-boundary bind + js interop) is proven.

_`spikes/` are evidence artifacts, isolated from the workspace/CI/wasm build; they'd be removed before any productization merge._


Faolain:
From memory @marcus-pousette when I looked into this a few months ago there were many regressions on the rust-libp2p side that didn't make this worth it with many outstanding bugs for simple connectivity that were already addressed by js-libp2p. Unsure if this still stands but worthwhile to investigate otherwise you may see significant regressions that may not make this worth the effort (everything from poor interop with other libp2p nodes, bugs in the different connection types and so forth)

Faolain:
As an addendum this would be specifically problematic when wanting to package peerbit into apps that would run server/client side where you want to leverage all the different connection types. I haven't seen what the maintenance of rust libp2p is like but I saw even big issues sitting for a very long time that were solved by js-libp2p ages ago. It seems like you're making this opt-in so as a user one isn't locked in but I'm not sure how high value this is vs other stability/bug fixes elsewhere. Just my 2 cents ofc.

peerbit-org:
Closing this draft as an archived research branch, not rejecting its findings. The shared native network core is already on master through #988, and the native-safe log error follow-up landed through #998. This head is now 609 commits behind, conflicts with current master, contains 13.6k lines of spike/benchmark material, and predates #1167’s 2.1 control-plane work. A product native transport should restart as focused PRs from current master with mixed-version, NAT/relay, packaging, and maintenance gates. Near-term native-index scaling remains tracked in #1308. We will retain this closed PR/branch as the evidence archive.

PR #1111, opened 2026-07-20
fix(program): harden terminal lifecycle cleanup

peerbit-org:
## Summary

- make close and drop completion proofs depend on captured, module-private lifecycle state rather than forgeable runtime properties
- recover exact owner and child relationships when failed opens roll back, while preserving compatibility with legacy programs that do not expose trusted terminal proof
- add hostile override, retry, clone, foreign-module, and exact-owner regression coverage

This hardens the local Program and Handler lifecycle protocol. It does not change the Peerbit network or wire protocol.

## Validation

- `pnpm --filter @peerbit/program build`
- full `@peerbit/program` suite: 241/241
- hostile lifecycle matrix: 17/17
- exact rollback matrix, repeated: 32/32
- lifecycle edge cases: 5/5
- executable hostile probes: 5/5
- foreign module-copy probe: 1/1
- downstream database suite: 166/166
- focused downstream owner/former-owner cases: 8/8
- Prettier and staged diff checks
- lockfile unchanged

## Known baseline issue

A workspace-wide build was also attempted, but stops in the unchanged `@peerbit/any-store-interface` package because `@dao-xyz/borsh` cannot be resolved in this worktree. The affected Program package build and all scoped/downstream suites above pass.


peerbit-org:
Closing after independent adversarial review. The goal here treats same-realm Program overrides as a hostile security boundary, but such code can replace lifecycle proof hooks or schedule close/drop mutations after its returned promise settles; no Handler protocol can make arbitrary in-process code untrusted. The current draft also introduced correctness, cross-copy compatibility, dynamic-dispatch, and scaling risks while trying to enforce that model. This is local lifecycle behavior, not a Peerbit wire/network vulnerability. We should document the trusted override contract (await or return the base terminal operation exactly once) and land any reproducible defect under that normal contract as a small, independently reviewed PR.

PR #1130, opened 2026-07-24
fix(shared-log): harden checked prune handoffs

peerbit-org:
## Summary

- replace legacy checked-prune authorization with capability-gated, request-correlated handoffs
- revalidate block presence, leadership, requester ownership, and exact confirmation quorums at the final grant/deletion boundaries
- prevent child handoffs from deleting ancestors and fail closed for legacy or stale peers
- bound vectors, hash lengths, retries, pending missing-entry work, handler/callback work, and concurrent request/grant delivery
- retain admission charges for deadline-abandoned non-cancellable work until it actually settles

## Validation

- `pnpm run build`
- focused checked-prune, admission, retry, pending-IHave, raw-sync, exchange-heads, replication, and sharding tests pass
- final focused aggregate: 72 passing; dedicated protocol/resource aggregate: 50 passing
- full shared-log run reached 2,152 passing / 9 pending; one untouched accumulated-load synthetic memory threshold failed and then passed immediately in isolation
- 100-entry receive benchmark (3 measured runs): native-confirm ~144k ops/s; pending-IHave ~207k ops/s
- `git diff --check`

## Release

Includes a patch changeset for `@peerbit/shared-log`.

### DIFF
--- a/.changeset/safe-checked-prune-handoffs.md
+++ b/.changeset/safe-checked-prune-handoffs.md
@@ -1,0 +1,5 @@
+---
+"@peerbit/shared-log": patch
+---
+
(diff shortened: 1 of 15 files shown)


peerbit-org:
Superseded by #1132, which rebuilds this work on a clean branch after the full adversarial and scope review and includes the final race fixes plus validation.

PR #1138, opened 2026-07-25
fix(pubsub): converge auto-root overlays

peerbit-org:
## Summary

- converge retained shard overlays when automatic root candidates change during late peer joins
- reject differing peer-query shard roots while auto mode is active, while preserving locally configured roots, resolvers, and trackers
- fence root resolution and fanout opening by candidate generation and control-plane lifecycle
- preserve overlapping reconciliation requests with a dirty-loop scheduler
- bind direct root-control messages to their verified transport peer and bind query responses to the peer actually queried

A leaf that intentionally consumes a queried gateway's internal shard roots must first call `setTopicRootCandidates([])`. Non-shard queried roots and locally configured authorities retain their existing behavior.

## Validation

- `@peerbit/pubsub` full Node suite: **163 passing**
- focused subscribe race/security/topology suite: **22 passing**
- changeset validation: patch release (`@peerbit/pubsub` 5.3.6)
- exact packed commit hot-patched into `distributed-backup-v2`:
  - deterministic 64 MiB payload, 256 chunks, three isolated peers, two receivers
  - source force-stopped with SIGKILL before receiver validation
  - both receivers passed local-only metadata, pointer, byte/chunk count, SHA-256, and CRC-32 validation
  - receiver sync wall times: 1.60 s and 1.65 s
- `git diff --check` clean

## Review notes

The production change is isolated to pubsub. Most added lines are targeted race, lifecycle, security, gateway-compatibility, and real three-node star regressions; no shared-log hardening changes are bundled here.

### DIFF
--- a/.changeset/calm-roots-converge.md
+++ b/.changeset/calm-roots-converge.md
@@ -1,0 +1,7 @@
+---
+"@peerbit/pubsub": patch
+---
+
+Converge auto-discovered shard roots across late peer joins and bind direct root-control messages to their verified transport peer.
+
+Leaves that use a queried gateway's internal shard roots must disable automatic candidates with `setTopicRootCandidates([])`; locally configured roots, resolvers, and trackers remain authoritative.
--- a/packages/transport/pubsub/src/index.ts
+++ b/packages/transport/pubsub/src/index.ts
@@ -230,6 +230,14 @@ export type TopicControlPlaneOptions = DirectStreamOptions & {
 	subscriberCacheMaxEntries?: number;
 };
 
+type EnsureFanoutChannelOptions = {
+	ephemeral?: boolean;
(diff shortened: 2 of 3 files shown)


peerbit-org:
Merge review complete as `peerbit-org`.

All 17 functional test shards, the workspace build, fanout gate, shared-log batch gate, and native/rust-core matrix passed. The two red security-contract jobs are a baseline audit finding also present on `master`: newly disclosed GHSA-mh99-v99m-4gvg in transitive React Native build-tooling copies of `brace-expansion`, unrelated to this dependency-free pubsub diff. The compatible v1/v2 upstream backports are still awaiting release, so remediation will stay in a separate dependency PR rather than being bundled here.

PR #1140, opened 2026-07-25
fix: gate raw sync requests by peer capability

peerbit-org:
## Summary

- gate raw Simple sync request variants per target on that peer’s advertised capability
- fail closed to legacy request variants when capability information is missing or unknown
- pass the existing capability state through Simple, Rateless, and custom synchronizer construction paths
- cover missing-capability and mixed raw/legacy target behavior

## Why

Raw receive integrity and admission checks are already enforced, but local raw opt-in could previously select bulk request variants before knowing whether an older remote could decode them. This closes that mixed-version compatibility gap before any downstream file-chunk opt-in. It does not add another wire format.

## Validation

- `pnpm run build`
- `pnpm run test:part-4` — 2,286 shared-log tests and 1 proxy test passing
- focused sync-chunking/raw suite — 104 passing
- downstream 64 MiB, three-node filesystem hotpatch benchmark — both receivers matched exact chunk count, byte count, SHA-256, and CRC-32; raw profiling was active; all tracking and repair targets returned to zero
- `git diff --check`
- independent read-only review: no actionable findings

## Notes

The package-wide lint command currently reports pre-existing formatting errors in untouched files on master, so this PR avoids unrelated mass formatting. Release PR #1139 remains separately held by the fail-closed dependency audit.

### DIFF
--- a/.changeset/safe-raw-peer-sync.md
+++ b/.changeset/safe-raw-peer-sync.md
@@ -1,0 +1,5 @@
+---
+"@peerbit/shared-log": patch
+---
+
+Gate raw Simple sync request variants on each remote peer's advertised capability so mixed-version peers fall back to legacy messages.
--- a/packages/programs/data/shared-log/src/index.ts
+++ b/packages/programs/data/shared-log/src/index.ts
@@ -16342,6 +16342,8 @@ export class SharedLog<
 				sync: options?.sync,
 				isEntryRecentlyKnownByPeer: (hash, peer, maxAgeMs) =>
 					this.isEntryRecentlyKnownByPeer(hash, peer, maxAgeMs),
+				peerSupportsRawExchangeHeads: (peer) =>
+					this.peerSupportsRawExchangeHeads(peer),
 				sendRawExchangeHeads,
 			});
 		} else {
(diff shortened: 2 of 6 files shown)


peerbit-org:
Bot merge review:

- exact reviewed head: `6e70e1a2ff5305f102021306ee77bbd1072d20cb`
- independent adversarial review: no correctness, security, lifecycle, compatibility, or scope findings
- Build Workspace, Fanout Gate, Shared Log Sync Batch Gate, Native, and every concrete `test:ci` partition passed
- local `test:part-4`: 2,286 shared-log tests plus proxy test passed
- the only failures are the Node 22/24 dependency contracts for `GHSA-mh99-v99m-4gvg` (`brace-expansion`, 28 paths), exactly matching the current master run: https://github.com/dao-xyz/peerbit/actions/runs/30144638552
- this PR changes no dependency or workflow files and does not waive or weaken the release security gate; release PR #1139 remains held until reviewed upstream v1/v2 fixes are published and recognized by the advisory

Merging this focused mixed-version protocol compatibility fix with the baseline audit exception documented here.

PR #1168, opened 2026-08-03
perf(shared-log): memoize leader plans between topology changes

peerbit-org:
Leader election runs 2–3 times per received entry and once per entry in every rebalance/prune sweep. On the TS tier each run costs two full replication-index scans plus 4–12 range-index iterator reads per cursor; even on the native planner tier each run pays leader-selection-context construction (including a sorted replication-index query inside `getDefaultMinRoleAge`, context-cached for only 50ms). The results are pure functions of the replication-range set, so repeated elections between topology changes recompute identical plans.

## Design

- New `LeaderPlanCache` (own module), consulted by `_findLeaders` (cursor path) and `_findLeadersFromHashGid` (gid path). Coherence is **structural, not temporal**: a private version bumps and the store wipes inside `invalidateLeaderSelectionContextCache` — the existing chokepoint fired on `replication:change`, `replicator:mature`, subscriber-cache invalidation — and an async fill is only stored when the version captured before the computation still holds (the rateless-iblt encoder-cache pattern). Plans are neither served nor stored while replication-range mutations are in flight, mirroring the receive path's ownership-revision rule.
- Only the two hot roleAge buckets cache (dynamic default, and 0 as used by all sweeps); candidate filters or other explicit roleAge bypass. Served and stored plans are cloned — callers mutate the maps and their value objects.
- **The checked-prune delete boundary bypasses the cache** via a new `freshLeaderPlan` option: its invariant is a fresh ownership decision inside the global mutation lane immediately before `log.remove`. An adversarial multi-lens review traced every deletion path: no delete consumes a cached plan (the only other `log.remove` is the public `prune({unchecked: true})` opt-in). Staleness elsewhere is one-directional — a cached plan can only briefly under-report leaders, the conservative direction for every consumer (missed grants retry; overshoot corrects via later sweeps).
- The review also surfaced two pre-existing invalidation-chokepoint gaps that a result cache would amplify — `_isReplicating` flips on `replicate()`/`unreplicate()` denial paths, and `uniqueReplicators` adds in `pruneOfflineReplicators` — both now fire the chokepoint. TTL is 50ms, deliberately: a range re-crossing maturity after the dynamic roleAge regrew fires no timer event, so expiry is the correctness mechanism for that edge and the bound matches the previous context-cache window.

## Where the wins land (honest map, from the review trace)

The native-backbone receive fast path uses whole-message batch planners and is untouched by this cache. The wins are: rebalance and prune sweeps (per-gid reuse within a sweep; entries sharing a gid share a key), the dynamic-default-roleAge bucket (skips the recurring sorted index query + subscriber lookup per election), gid-reference merging on the append-delivery path, and the entire TS fallback tier (browsers), where each avoided election saves two full index scans plus per-cursor iterator reads.

## Validation

- Fresh-worktree local runs (fresh install, mock session): `isLeader` 21, `waitForReplicator` 18, `strict live-range leader routing` 1, `receive admission` 16, `checked prune` 45, `replicate` 28, `observer` 3 — all passing, exit codes verified individually.
- New unit spec for the cache class (clone isolation, version-guarded fills, invalidation, TTL, size bound) runs in shard part-6 under the existing `isLeader` grep.
- 4-lens adversarial review (staleness safety, invalidation completeness, implementation details, perf reality): no blockers; key-collision, lifecycle-contract, clone-depth, and cache-semantics checks all verified clean.
- PR CI runs the full shared-log matrix (part-4/6/7a–7e) plus the sync-batch hang gate; base-vs-head benchmark comparison and chaos-seed sims dispatched separately.

### DIFF
--- a/packages/programs/data/shared-log/src/index.ts
+++ b/packages/programs/data/shared-log/src/index.ts
@@ -186,6 +186,7 @@ import {
 	type Numbers,
 	createNumbers,
 } from "./integers.js";
+import { LeaderPlanCache } from "./leader-plan-cache.js";
 import { TransportMessage } from "./message.js";
 import { PIDReplicationController } from "./pid.js";
 import {
@@ -2638,6 +2639,14 @@ const RECALCULATE_PARTICIPATION_MIN_RELATIVE_CHANGE_WITH_MEMORY_LIMIT = 0.001;
 const RECALCULATE_PARTICIPATION_RELATIVE_DENOMINATOR_FLOOR = 1e-3;
 const TOPIC_SUBSCRIBERS_CACHE_TTL_MS = 250;
 const LEADER_SELECTION_CONTEXT_CACHE_TTL_MS = 50;
+// Leader-plan results are invalidated structurally (replication:change,
+// replicator:mature, subscriber changes). The TTL bounds wall-clock maturity
+// flips that have no timer at all — a range re-crossing maturity after the
+// dynamic default roleAge regrew is only visible via expiry — so it is
+// load-bearing for that edge and must stay at the same order as the
+// leader-selection context TTL.
+const LEADER_PLAN_CACHE_TTL_MS = 50;
+const LEADER_PLAN_CACHE_MAX = 10_000;
 const ADAPTIVE_REBALANCE_IDLE_INTERVAL_MULTIPLIER = 5;
 const ADAPTIVE_REBALANCE_MIN_IDLE_AFTER_LOCAL_APPEND_MS = 10_000;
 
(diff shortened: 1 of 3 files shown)


peerbit-org:
Benchmark validation record (both `benchmarks.yml` dispatch failures investigated and cleared):

- **Document benchmark crashes** ([run 30832555089](https://github.com/dao-xyz/peerbit/actions/runs/30832555089) head, [run 30835613655](https://github.com/dao-xyz/peerbit/actions/runs/30835613655) base): identical `NotStartedError` in `Documents.handleChanges` → `SQLiteIndex.assertOpen` on **both head and unmodified master** — a pre-existing close-ordering race in the document store (an in-flight `joinRecursively`'s change events reach the index after close), unrelated to this PR. Locally (Node 22, fresh installs): 3/3 file-ingest passes on head *and* base. Tracked as a separate fix.
- **Apparent +9–12% sync-catchup sweep regression in run 30832555089**: measured as noise. Two CI *base* runs of identical code differ by −6% to −10% (the same magnitude), and `requestMaybeSyncTotal` swings 30–70% between identical-code runs. Local alternated A/B (3× per side, same machine, Node 22): config 512 → head **−5.5%** (sign reversal vs the single CI sample), config 16384 → −0.9%; counters statistically identical. The other suites (StartSync decoder, sender StartSync, PID model) were neutral in CI (±3.5% noise floor).

The first dispatch also surfaced a missing `rustup target add` in the benchmark workflow, fixed in #1169.

PR #1181, opened 2026-08-04
refactor(shared-log): stage-4.5 PR-1 — extract the coordinate-persistence module

peerbit-org:
First of the two final stage-4 PRs. The shared coordinate-persistence infrastructure — 39 methods and four state fields used by both the append and receive paths — leaves index.ts for `src/coordinate-persistence.ts` (index.ts: 26,109 → 24,836 lines). The planning pass explicitly scoped **out** full onMessage decomposition: the fused Raw/ExchangeHeads unit (1,870 of the hottest lines in the repo) stays untouched, deferred past the stage-5 fence collapse that would otherwise churn it twice.

## Structure (three commits, pins first)

1. **Nine pins** (`coordinate-persistence-pins.spec.ts`): flush predicate/watermark, generation-checked durable rollback, forget-before-delete ordering, backbone-only batch atomicity — green against unmoved code first.
2. **State ownership**: `CoordinatePersistenceCoordinator` owns the four fields under historical names; 71 reader sites become direct property hops (no getters in hot loops); the plan's blanket `reset()` was **rejected by the implementer** — the three per-open sites reset *different* field subsets (the resident mirror must survive `openNativeBackbone`; the generation ratchet survives everything), so each keeps its exact legacy writes. Review verified subset-for-subset.
3. **Verbatim moves**: all 39 methods byte-identical under the documented glue mapping — proven by diff at implementation time and **independently re-verified in review** (automated sweep + hand-diff of the 8 most complex). 29 KEEP-OLD delegators keep every external call site untouched; no runtime import cycle (compiled module strictly downstream, type-only imports erased).

## Review

One confirmed finding — the same class the discipline caught in the controller PR: code correct, **pin toothless**. P2's stale-generation clause couldn't distinguish the guard from an always-rollback because the stale snapshot captured identical state (the reviewer proved the mutant survives via a load-time hook after catching their own false-negative from mocha's config respawn). Fixed with a visible-state discriminator — the row is deleted under the newer generation, so an unconditional stale rollback would resurrect it — and the mutant now dies on exactly that assertion, re-verified both directions.

## Validation

25-suite sweep (widest of the refactor) 0 failing; both chaos suites green; per-commit gates green throughout; ratchet 13 entries with `_nativeCoordinateMutationGenerations` moved file-to-file; 21 test spy sites re-pointed with interception multiplicity empirically preserved.

Remaining after this: stage-4.5 PR-2 (receive dispatcher + five cold control-plane branches) and the sync-module unification.


peerbit-org:
part-2a failure diagnosed and fixed in the amended head: four document-package tests stub `canUseNativeBackboneResidentCoordinateState` by name on the SharedLog instance to force the direct-fallback path, and the extraction's two coordinator-internal calls bypassed the host stub — the exact host-routed-dispatch rule, missed because the spy census covered only shared-log's own tests. The predicate's internal calls now route through a host-routed deps closure (documented at the deps interface). Verified: the four failing tests green locally (68 passing across both durable-native document suites), shared-log regression subset green (105 tests). The repo-wide census found the other three externally-stubbed moved names are delegator-covered and unaffected.

PR #1235, opened 2026-08-13
fix(deps): move @libp2p/webrtc to a prebuilt node-datachannel

peerbit-org:
Fixes the recurring Node 24 install failures at their source, rather than working around them.

## Root cause

The workspace resolved `@libp2p/webrtc@6.0.15`, which requires `node-datachannel ^0.29.0`. **`node-datachannel@0.29.0` publishes no prebuilt binaries at all** (zero release assets), so `prebuild-install` can never succeed and every install compiles it from source. That source build is broken on Node 24: `prebuild`/`prebuild-install` are deprecated and their `napi-build-utils` cannot resolve NAPI v10, aborting with `does not support N-API version undefined` — upstream https://github.com/murat-dogan/node-datachannel/issues/428, open since July, no release since April.

CI mostly hid this because pnpm's side-effects cache holds a previously-compiled binary. On any cache miss, Node 24 jobs fail — which is what happened on #1234, where the failure appeared under the *pinned* 24.18 runner, proving this was never specific to 24.19.

## The fix

`@libp2p/webrtc@6.0.29` declares `node-datachannel ^0.32.3`, and 0.32.3 ships `napi-v8` prebuilds for every supported platform. Since N-API is ABI-stable those binaries are forward-compatible, so installs need neither a compiler nor the broken build tooling. Verified directly: `npm install node-datachannel@0.32.3` succeeds on Node 24 (exit 0, prebuilt binary used), while 0.29.0 has no asset to fetch.

The declared ranges were already `^6.0.15`, so 6.0.29 was always permitted — the lockfile was simply stale. Ranges are bumped to `^6.0.29` in the four manifests so a future resolution cannot silently fall back to a prebuild-less line.

## Consequences

- Node 24 installs stop depending on a warm pnpm cache.
- Installs get faster and stop requiring cmake/a toolchain for this dependency.
- Once this lands and CI is green across a few runs, the `24.18.x` pin from #1231 can likely be reverted to floating `24.x` — deliberately left for a follow-up so this change is judged on its own.

Patch changesets for the three published packages whose dependency range moved; `stream-browser-test-vite` is private.

### DIFF
--- a/.changeset/webrtc-node-datachannel-prebuilds.md
+++ b/.changeset/webrtc-node-datachannel-prebuilds.md
@@ -1,0 +1,9 @@
+---
+"@peerbit/libp2p-test-utils": patch
+"@peerbit/react": patch
+"peerbit": patch
+---
+
+Move `@libp2p/webrtc` to `^6.0.29`, which depends on `node-datachannel ^0.32.3` instead of `^0.29.0`.
+
+`node-datachannel@0.29.0` publishes no prebuilt binaries at all, so every install compiled it from source. That source build fails on Node 24 — `prebuild`/`prebuild-install` are deprecated and their `napi-build-utils` cannot resolve NAPI v10, aborting with "does not support N-API version undefined" (upstream: https://github.com/murat-dogan/node-datachannel/issues/428). `0.32.3` ships napi-v8 prebuilds for every supported platform, and N-API is ABI-stable, so installs no longer need a compiler or the broken build tooling.
--- a/packages/clients/peerbit-react/package.json
+++ b/packages/clients/peerbit-react/package.json
@@ -53,7 +53,7 @@
 		"@libp2p/circuit-relay-v2": "^4.1.7",
 		"@libp2p/crypto": "^5.1.14",
 		"@libp2p/interface": "^3.1.1",
-		"@libp2p/webrtc": "^6.0.15",
+		"@libp2p/webrtc": "^6.0.29",
 		"@libp2p/websockets": "^10.1.7",
 		"@multiformats/multiaddr": "^13.0.1",
 		"@peerbit/canonical-client": "workspace:*",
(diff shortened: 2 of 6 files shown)


peerbit-org:
Closing: the fix is correct in principle but cannot land as a scoped dependency bump. Recording what was established so this is not re-investigated from scratch.

**Root cause (confirmed).** `node-datachannel@0.29.0` publishes **zero** prebuilt binaries, so every install compiles it from source, and that source build is broken on Node 24 — `prebuild`'s `napi-build-utils` cannot resolve NAPI v10 (upstream https://github.com/murat-dogan/node-datachannel/issues/428). CI has been surviving on pnpm's side-effects cache; **any cache miss fails**, and it fails on the pinned 24.18 runner too, so this was never a 24.19 problem and the #1231 pin does not actually mitigate it.

**Why the scoped bump does not work.** The first `@libp2p/webrtc` with a prebuilt `node-datachannel` is 6.0.21, which requires `@libp2p/interface ^3.2.2`. Bumping webrtc alone gives two copies of `@libp2p/interface` and `Transport`'s branded symbol stops matching (`TS2322` in `libp2p-test-utils/src/transports.ts`). Deduping onto 3.2.5 is legal — every workspace package already declares `^3.1.1` — and actually shrinks the lockfile, but then `session.ts` fails on `Libp2pWithServices` and `StreamMuxerFactory`: the rest of the stack (libp2p core, muxers) is still built against 3.1.x. Verified locally with a full workspace build, not guessed.

**Options, for whoever picks this up.** (1) Upgrade the whole libp2p stack — the real fix, but a project with genuine blast radius on the transport layer, not a drive-by. (2) Override `node-datachannel` to 0.32.3 while keeping webrtc 6.0.15 — bypasses the declared `^0.29.0`, so it needs the WebRTC transport tests to prove the binding API did not drift. (3) Drop the Node 24 entry from the security matrix and accept losing that forward-compat signal.

Nothing here is urgent: the failure is intermittent and cache-dependent, and no published artifact is affected.

PR #1261, opened 2026-08-14
ci: lint every source root and make the warning budget a ratchet

peerbit-org:
`pnpm lint` enumerated `packages` only, so **147 source files were never linted** — every `.ts`/`.tsx` of the shipped website under `apps/`, the docs examples, `tools/`, and all of `scripts/` **including the CI guard scripts themselves**.

And the budget that looked like a ratchet was decoration:

```js
if (errorCount > 0 || warningCount > 9_999) {
```

The actual warning count is **15**. At 9,999 every `warn`-level rule in `eslint.config.js` was structurally non-blocking — unused vars and the rest could accumulate without limit and CI stayed green.

## Widening was nearly free, but not entirely

All four new roots were clean except for **6 errors**, in files nothing had ever checked:

**`scripts/ci/check-changeset-required.mjs`** ×3 — `no-regex-spaces` in the regexes that parse workflow YAML by indentation. Exactly the case the rule exists for ("Spaces are hard to count"):

```diff
-/^  ([a-z_][a-z0-9_-]*):$/gm        +/^ {2}([a-z_][a-z0-9_-]*):$/gm
-/^      - name: /gm                 +/^ {6}- name: /gm
-/^        uses: ([^\n]+)$/gm        +/^ {8}uses: ([^\n]+)$/gm
```

Equivalent, and now the intent is readable. This is the repo's largest guard, so its 2,604-line self-test was re-run: **52/52 pass**.

**`apps/peerbit-org/scripts/docsContentPlugin.ts`** ×2 — `\"` inside a template literal, where no escape is needed. Output-identical (it builds RSS XML, so that mattered to check).

**`scripts/file-share/templates/upload-benchmark.local.e2e.spec.ts`** ×1 — `no-unsafe-finally`. This one is a **false positive and the code is correct**, so it gets a documented suppression rather than a restructure: a `throw` in `finally` normally discards the in-flight exception, but the catch above assigns `benchmarkFailure = error` as its *first* statement (line 3486) and the `finally` folds that value into the `AggregateError`, so the original benchmark failure is reported alongside the cleanup failures instead of replacing them. The rule sees the syntactic shape, not the capture.

## The ratchet

`MAX_WARNINGS = 15` — the real count — with a message naming the overage, and a comment that it may go DOWN freely while raising it must be a deliberate visible edit.

Mutation-verified: adding one unused variable gives

```
✖ 16 problems (0 errors, 16 warnings)
lint: 16 warnings exceeds the ratchet of 15.
rc=1
```

and removing it returns to rc=0.

No workflow change needed — the Lint step already runs `pnpm run lint`; it just now sees the whole repo.

### DIFF
--- a/apps/peerbit-org/scripts/docsContentPlugin.ts
+++ b/apps/peerbit-org/scripts/docsContentPlugin.ts
@@ -134,7 +134,7 @@ function buildRssFeed({
 				"<item>",
 				`<title>${xmlEscape(i.title)}</title>`,
 				`<link>${xmlEscape(link)}</link>`,
-				`<guid isPermaLink=\"true\">${xmlEscape(link)}</guid>`,
+				`<guid isPermaLink="true">${xmlEscape(link)}</guid>`,
 				`<pubDate>${xmlEscape(pubDate)}</pubDate>`,
 				desc ? `<description>${desc}</description>` : "",
 				"</item>",
--- a/scripts/ci/check-changeset-required.mjs
+++ b/scripts/ci/check-changeset-required.mjs
@@ -1710,7 +1710,7 @@ export const validateChangesetGuardWorkflow = (
 		);
 	}
 	const jobsSource = source.slice(source.indexOf("\njobs:\n") + 7);
-	const jobNames = [...jobsSource.matchAll(/^ {2}([a-z_][a-z0-9_-]*):$/gm)].map(
+	const jobNames = [...jobsSource.matchAll(/^  ([a-z_][a-z0-9_-]*):$/gm)].map(
 		(match) => match[1],
 	);
 	if (!isDeepStrictEqual(jobNames, ["changeset_guard"])) {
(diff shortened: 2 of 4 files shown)


peerbit-org:
## Corrected: the changeset guard is a frozen root-of-trust boundary

The first push failed `Enforce Changeset Coverage` and `Build Workspace`, both with:

```
scripts/ci/check-changeset-required.mjs is a frozen executable root-of-trust
boundary and must remain byte-identical to the trusted base
```

**That rejection is correct and I have backed the change out.** The guard runs under `pull_request_target` with elevated permissions, so it fails any PR whose copy of itself differs from the trusted base — which is precisely what stops a PR from editing the check that decides whether that PR is safe. My eslint autofix was semantically equivalent (`/^  (…)/` → `/^ {2}(…)/`) but an edit is an edit, and the boundary does not care about intent.

Note the consequence: that file and its test **cannot be changed through a PR at all** — only by a direct push to master. I have not done that and would not without you asking; bypassing the control to satisfy a cosmetic lint rule would be the wrong trade.

So both files are excluded from the lint scope, byte-identical to master again (verified), with the reasoning recorded at the exclusion. Their real guard remains the 2,604-line self-test, which still runs.

The other three lint fixes stand — they are in ordinary files:
- `apps/peerbit-org/scripts/docsContentPlugin.ts` — redundant `\"` in a template literal (output-identical RSS)
- `scripts/file-share/templates/upload-benchmark.local.e2e.spec.ts` — documented `no-unsafe-finally` suppression, code verified correct

`pnpm run lint` is rc=0 with 0 errors / 15 warnings, and the ratchet still fires on a 16th.

PR #1263, opened 2026-08-14
chore(deps): refresh the libp2p stack and drop the node-datachannel override

peerbit-org:
PR #1235 concluded that moving off `@libp2p/webrtc@6.0.15` needed a coordinated stack upgrade. It needed a lockfile refresh.

## What was actually wrong

Every libp2p package sat at **exactly the floor of its own caret range** — the lockfile was written when those were latest and never refreshed. The manifests already permitted the current versions:

| package | lockfile | range | now |
|---|---|---|---|
| `libp2p` | 3.1.7 | ^3.1.7 | **3.3.8** |
| `@libp2p/interface` | 3.1.1 | ^3.1.1 | **3.2.5** |
| `@libp2p/interface-internal` | 3.0.14 | ^3.0.14 | **3.1.11** |
| `@libp2p/webrtc` | 6.0.15 | ^6.0.15 | **6.0.29** |
| `@libp2p/tcp` | 11.0.14 | ^11.0.14 | **11.0.26** |
| `@libp2p/identify` | 4.0.14 | ^4.0.14 | **4.1.12** |
| `@libp2p/circuit-relay-v2` | 4.1.7 | ^4.1.7 | **4.2.11** |

#1235 failed because it bumped **webrtc alone**: 6.0.21+ wants `@libp2p/interface ^3.2.2` while the lockfile held everything else at 3.1.1, so two copies of `@libp2p/interface` coexisted and `Transport`'s branded symbol stopped unifying (`TS2322`). Updating the family in **one resolution** lands them all on interface 3.2.5 — one copy, no branded-type clash.

## Every failure was the same mechanism

Each fix revealed the next duplicated type-carrying package:

1. **`@libp2p/interface`** 3.1.1 + 3.2.5 → 3.2.5 (family update)
2. **`@multiformats/multiaddr`** 13.0.1 + 13.0.3 → 13.0.3 — libp2p 3.3.8 wants ^13.0.3; two `Multiaddr` types meant `Libp2pWithServices` stopped satisfying a duck-typed helper in `libp2p-test-utils`
3. **`uint8arraylist`** 2.4.8 + 3.0.2 → v3 in 5 manifests. A **major** boundary, so `pnpm dedupe` could not cross it; v3 made `Uint8ArrayList` generic
4. **`it-length-prefixed`** 10 + 11 → v11 in 2 manifests (v10 pulls uint8arraylist v2)
5. **`interface-datastore`** 9.0.2 + 10.0.1 → override to 10.0.1

On (5): `@dao-xyz/datastore-level@11.2.0` — our own fork — pins `interface-datastore ^9.0.2`, so an override is the only way to dedupe without republishing it. That is safe here because **v9 and v10 export exactly one runtime symbol, `Key`**; the entire major is type-level. Republishing the fork against v10 is the cleaner end state.

## The payoff

The `node-datachannel` override is **removed**. It existed to force a prebuilt 0.32.3 because webrtc 6.0.15 pulled the prebuild-less 0.29.0. webrtc 6.0.29 requires `^0.32.3` itself, so the resolution is 0.32.3 with no override at all — verified after removal. Our dev install now matches what consumers actually resolve.

## Validation

```
pnpm run build                          rc=0, 0 errors
@peerbit/blocks                         69 passing
@peerbit/pubsub                        218 passing
peerbit                                109 passing
```

**One caveat, stated plainly:** `@peerbit/stream` runs 46 of its ~187 defined tests and reports no summary while exiting 0. I verified this is **identical on master** (master CI part-3 shows the same 46 ✔ and the same final test), so it is pre-existing and not caused by this change — but it means CI's signal on the transport package is weaker than it appears. Filed separately; it deserves its own fix.

A stale `ci.yml` comment describing the removed override is corrected in the same commit.


peerbit-org:
## Fixed both CI failures

**1. `@peerbit/stream` test build** — `Uint8ArrayList<ArrayBufferLike>` vs `Uint8ArrayList`. A v2 copy still reached the *tests*: `@chainsafe/libp2p-noise@17.0.0` and `@chainsafe/libp2p-yamux@8.0.1` — both **latest** — still depend on `uint8arraylist@2`, and stream's tests meet them through the test session. So that upstream lag is real, it just surfaces at the test boundary rather than in src.

Resolved with a `pnpm.overrides` pin to 3.0.2, on the same evidence used for `interface-datastore`: **v2 and v3 export the identical runtime surface** (`Uint8ArrayList`, `isUint8ArrayList`), so the major is type-level and forcing it is runtime-safe for noise/yamux.

Verified with the exact command CI failed on — `aegir build --no-bundle` in `packages/transport/stream` — now rc=0, and the full workspace build is rc=0.

**2. Changeset coverage** — correctly required: bumping dependencies of published packages is release-relevant. Added a patch changeset covering all 20 packages the guard named.

## Standing caveat, unchanged

`@peerbit/stream` runs **46 of its ~187 defined tests** and exits 0 with no summary, stopping two tests into `messages-signing.spec.ts` and never reaching the 102 in `stream.spec.ts`. I confirmed this is **identical on master** (same 46 ✔, same final test), so it is pre-existing and not introduced here — but it does mean CI's signal on this package is roughly a quarter of what it appears to be, and this PR's transport validation is correspondingly weaker. Worth fixing on its own.

PR #1270, opened 2026-08-14
perf(shared-log): compute tick participation exactly instead of sampling

peerbit-org:
The 1 Hz adaptive rebalance tick issued **27 index round-trips per second per open log**. 25 of them came from one call, and an exact single-query equivalent was already sitting in the same function.

## The redundancy

```ts
async calculateTotalParticipation(options?: { sum?: boolean }) {
    if (options?.sum) {
        const ranges = await this.replicationIndex.iterate().all();   // 1 query
        let sum = 0;
        for (const range of ranges) sum += range.value.widthNormalized;
        return sum;
    }
    return appromixateCoverage({ ..., samples: 25 });                 // 25 queries
}
```

The tick called it **without** `{ sum: true }`, taking the sampling path: 25 sequential `count()` queries over a fixed 25-point grid, plus ~175 freshly-allocated query AST objects, every second.

The two branches compute the same quantity. Mean coverage of the domain **is** the sum of the range widths — `appromixateCoverage` is a Monte-Carlo estimate of exactly what `{ sum: true }` computes directly. So the tick was paying 25 queries for a noisy approximation of a value it could read exactly in one.

Tick cost goes **27 → 3** index round-trips. This is not an opt-in path: `replicate: true` and `replicate: undefined` both normalise to adaptive, so it applies to every default-configured log.

## This is a behaviour change, and was treated as one

The value feeds the replication controller's `totalFactor`, and the call site carries a standing `// TODO use this._totalParticipation when flakiness is fixed` — someone has been bitten changing this input before. The controller now sees a stable value where it previously saw a sampled one. Strictly better input, but different input.

So it was validated against the convergence suites rather than on the query count, and run **twice** to distinguish a stable controller from a lucky run:

| suite | pass 1 | pass 2 |
|---|---|---|
| `u32-simple sharding` | 31 passing | 31 passing |
| `u32-simple replication` | 64 passing | 64 passing |
| `PIDReplicationController` + `ranges:` | 1514 passing | — |

The convergence diagnostics are the substantive evidence. Three peers at full coverage settle to a participation of ~1/3 each, both times:

```
pass 1:  0.33350  0.33422  0.33476
pass 2:  0.33160  0.33546  0.33577
```

That is the controller reaching the correct steady state on the exact input, not merely a suite going green.

## Note on what is not claimed

No wall-clock benchmark is quoted. The census that surfaced this also found that the four base-vs-head harnesses have no documented noise floor and that CI runs the sync-peer-state comparator's unit tests but never the benchmark itself — so there is currently no trustworthy way to attach a millisecond figure to this. The defensible claim is the one made above: 25 fewer index queries per tick, derived from the code, with convergence unchanged.

### DIFF
--- a/.changeset/tick-exact-participation.md
+++ b/.changeset/tick-exact-participation.md
@@ -1,0 +1,19 @@
+---
+"@peerbit/shared-log": patch
+---
+
+Compute total participation exactly instead of sampling it 25 times per second.
+
+The adaptive rebalance tick called `calculateTotalParticipation()` on its default
+path, which Monte-Carlo estimates coverage with `appromixateCoverage({ samples:
+25 })` — 25 sequential index `count()` queries per tick. The same function
+already offers `{ sum: true }`, which totals `widthNormalized` across the
+replication index in a single `iterate().all()`.
+
+The two compute the same quantity: the mean coverage of the domain is the sum of
+the range widths. The sampled form was an estimate of the exact one, so this
+removes 25 of the tick's 27 index round-trips per open adaptive log — which is
+every default-configured log, since `replicate: true` and `replicate: undefined`
+both resolve to adaptive.
+
+The replication controller now sees a stable input rather than a sampled one.
--- a/packages/programs/data/shared-log/src/index.ts
+++ b/packages/programs/data/shared-log/src/index.ts
@@ -25014,7 +25014,22 @@ export class SharedLog<
 
 				const peersSize = (await peers.getSize()) || 1;
 				if (!isCurrent()) return false;
-				const totalParticipation = await this.calculateTotalParticipation();
+				// `{ sum: true }` totals widthNormalized over the replication
+				// index in ONE query. The default path instead calls
+				// appromixateCoverage({ samples: 25 }), which Monte-Carlo
+				// estimates the same quantity -- the mean coverage of the
+				// domain IS the sum of the range widths -- using 25 sequential
+				// count() queries. So this is the exact value at 1/25th of the
+				// index traffic, and it is 25 of the 27 queries this 1 Hz tick
+				// issued per open adaptive log.
+				//
+				// The controller consequently sees a stable input instead of a
+				// sampled one; that is a behaviour change, not just a speedup,
+				// which is why it is landing against the convergence suites
+				// rather than on the strength of the query count alone.
+				const totalParticipation = await this.calculateTotalParticipation({
+					sum: true,
+				});
 				if (!isCurrent()) return false;
 
 				const newFactor = this.replicationController.step({


peerbit-org:
Closing this. The change is theoretically sound and empirically wrong, which is worth recording.

## What it did

The 1 Hz adaptive tick took `appromixateCoverage({ samples: 25 })` — 25 sequential index `count()` queries estimating a quantity that `calculateTotalParticipation({ sum: true })` computes exactly in one query. Mean coverage of the domain **is** the sum of the range widths, so the sampled form is a Monte-Carlo estimate of the exact one. Swapping them takes the tick from 27 index round-trips per second to 3, on every default-configured log.

## Why it cannot land as-is

Two independent replication tests fail, and only one of them can be explained away:

**`replicate > replicator > dynamic by default`** — `AssertionError: expected 3 to be above 3`. `checkRoleIsDynamic` counts `replication:change` events and requires more than three. The tick only announces when `relativeDifference > minRelativeChange`, so with a sampled input the factor jittered across that threshold one extra time; with an exact input it converges directly and emits exactly three. That test was passing partly *because* of measurement noise.

**`u64-iblt > 2 peers write while joining`** — a convergence test in the u64 domain. This one is not an artefact of an assertion counting jitter; it is write-while-joining convergence behaving differently.

One failure would be a test encoding noise. Two, in different domains and different scenarios, means the controller's current tuning depends on the variance of its input. Removing that variance is a retuning problem, not a query-count optimisation.

This is exactly what the standing comment at the call site warns about:

```ts
totalFactor: totalParticipation, // TODO use this._totalParticipation when flakiness is fixed
```

Someone has been here before.

## What I deliberately did not do

Adjust `greaterThan(3)` to `greaterThan(2)`. The change would then have "passed", and peerbit would have shipped a controller whose convergence signature changed, with the evidence of that change edited away. If the exact input is desirable — and I think it is — the work is to retune against it and update the expectations knowingly, not to relax a threshold until CI agrees.

## What is worth keeping from this

- The 25-query sample and the 1-query sum are mathematically the same quantity; the tick is paying 25 queries per second per log for an estimate of a value it can read exactly.
- The controller's observable behaviour depends on that estimate's noise. Any future attempt needs the convergence suites **and** `replicate.spec.ts` **and** the u64 legs, not just the u32 ones — my local validation ran `^u32-simple sharding` and `^u32-simple replication` and missed both failures.
- A prerequisite worth doing first: the four base-vs-head benchmark harnesses still have no documented noise floor, so there is currently no way to show that a retune improved anything.

peerbit-org:
## Correction: measured the estimator, and my stated cause was wrong

I claimed above that the sampled path was noisy and that the failing tests were "passing because of measurement noise". Both wrong. Instrumenting the tick to log the sampled and exact values side by side gives a precise mechanism.

A diverging tick, with the actual row bounds:

```
sampled=1.040000  exact=1.000443
row A: w=0.508124  s1=2^63  e1=MAX_U64  s2=0  e2=149860821277935617
row B: w=0.492319  s1=0     e1=9081676042160525312
```

Row A is a wrap-around range — `[2^63, MAX] ∪ [0, 0.008124)`. It genuinely overlaps row B on `[0, 0.008124)`, so the extra hit is **real coverage, correctly counted**. A covers 13 of the 25 grid points, B covers 13, total 26 → 26/25 = 1.04. The estimator's arithmetic is right.

**The real defect is grid anchoring.** `getEvenlySpacedU32(from = zero, …)` places the first grid point at exactly 0, and a wrap-around range always stores its tail as `[0, ε)`. So `start2 <= 0 AND end2 > 0` holds for *any* ε > 0: a wrapping range always captures grid point 0 and always earns a full 1/25 = 0.04, however small its true tail. Here a 0.008-wide sliver was charged 0.04 — five times its weight.

That is a systematic upward bias, structurally guaranteed whenever a range wraps, which is common since ranges are placed by hash offset. It explains the observed sign: deltas ran overwhelmingly negative (sampled reading high) with two peers where one range wrapped, and mixed-sign with a single non-wrapping peer.

Two consequences for anyone picking this up:

1. There is no off-by-one to fix. `createContainingPointQuery` is already half-open (`start <= p AND end > p`), so a shared boundary between adjacent ranges is counted once. I had proposed "make the boundary half-open" as the fix; it would have been a no-op.
2. Removing the bias — by rotating the grid offset, or by using the exact sum as this PR did — moves the controller's equilibrium. That is why the two tests failed, and it is a retune, not a patch.

The estimator is coarse by design (`appromixateCoverage`), but the bias is not symmetric, and the controller has been tuned against the biased value.

PR #1288, opened 2026-08-15
bench: add shared-log resident and persistent scale census

peerbit-org:
## Summary

- Add an isolated-process shared-log scale census for resident native state and real persistent reopen paths.
- Measure memory, disk, close, and startup costs for chain graphs, high-frontier root graphs, and coordinate indexes.
- Checkpoint long runs atomically so a timeout or failure preserves completed rows and identifies the censored row.
- Add tiny end-to-end persistence coverage to ordinary CI and keep canonical 100k/1M profiles in the manual re-census workflow.
- Document the bounded-state contract needed for the 100M-global-entry goal in #1286.

## Why

The 100M target is a globally sharded deployment target, not permission for an ordinary peer to reconstruct all history. Existing resident measurements omitted persisted Rust indexes and the fresh startup path. Those paths matter because current indexes deserialize locally retained rows on open and the log rebuilds native graph state from retained shallow rows.

This PR establishes measurement truth; it does not claim the current stack already supports 100M entries or that one million locally materialized rows is healthy.

## Evidence

[Re-census run 31914725666](https://github.com/dao-xyz/peerbit/actions/runs/31914725666) measured resident state at 100k and 1M rows. A local persistent run then showed that a compact 1M chain snapshot could still take more than 30 minutes and roughly 4.3 GiB partial RSS to reopen.

[Independent-job run 31970105063](https://github.com/dao-xyz/peerbit/actions/runs/31970105063) let the 1M chain and roots cases complete separately:

| Scenario | Reopen time | RSS delta | Max RSS | Disk |
| --- | ---: | ---: | ---: | ---: |
| 1M chain | 2h 04m | 5.75 GiB | 6.58 GiB | 230 MiB |
| 1M roots | 1h 36m | 5.02 GiB | 5.68 GiB | 170 MiB |

The 1M persistent-coordinate reopen failed after roughly 2h10m when `RustIndex.init()` triggered wasm-bindgen's recursive-borrow/unsafe-aliasing guard. That existing indexer defect is tracked in #1308. The harness exposed it; this PR does not claim to have introduced it.

## Validation

After rebasing onto released master `a6580ceda`:

- 43 benchmark/harness contract tests passed.
- The full `@peerbit/shared-log` dependency build passed, including Rust/Wasm.
- The real persistent snapshot/reopen integration test passed.
- All six census scenarios passed at smoke scale.
- A mock-network SharedLog sync profile passed.
- Full current-head PR CI is the final merge gate.

[...]

Issue #1286, opened 2026-08-15
Scale shared-log to 100M global entries with bounded per-peer state


peerbit-org:
## Goal

Demonstrate that a shared-log deployment can retain and serve 100 million **global** entries while each ordinary storage peer keeps bounded memory, startup time, disk state, sync work, and control-plane state. This is not a requirement for one peer to materialize the complete 100M-entry history in RAM.

## Current evidence

- A synthetic native `LogGraphIndex` chain with realistic CID-length hashes used about 1,040 MiB additional RSS at 1M entries (about 1.09 KiB/entry) on one development machine. Linear extrapolation is roughly 109 GB at 100M entries for graph metadata alone, before blocks, coordinate state, the application index, and process overhead. This is a directional measurement, not a production capacity claim.
- The native graph is rebuilt by iterating every locally retained shallow-log row on open. Startup and graph memory are therefore O(local retained entries).
- The persisted Rust query index reconstructs its values and indexes in memory on open, so it is also O(local indexed rows).
- Shared-log frontier hydration, repair planning, range counts, and prune scans are O(local frontier rows). These can remain reasonable only if the local frontier is bounded.
- A recent 20k-missing-entry sync profile with many independent heads converged at about 390 entries/s; native receive persistence itself reached roughly 8k-12k entries/s. The dominant gap is sync/control orchestration, not raw mutation throughput.
- Placement hashes `entry.meta.gid`; one DAG/component is indivisible across its replica set. A hot video/channel/component therefore needs explicit shard granularity or a partitioning change.

## Define the scale contract first

Record and gate the following inputs rather than using only a global entry count:

- total historical entries and total payload bytes
- active storage peers versus non-storage viewers
- replication factor and required availability
- maximum and percentile locally retained entries, frontier rows, ranges, and gids
- maximum gid/component size and write rate
- steady-state writes/s and bytes/s
- peer churn and ownership-transfer rate
- restart memory/time and catch-up/rebalance SLOs
- hot/cold retention and media-blob storage model

Reference capacity examples at RF=2, before skew and overhead:

- 10k storage peers: 20k entry replicas per peer on average
- 100k storage peers: 2k entry replicas per peer on average

## Workstreams

- [ ] Add reproducible JSON census benchmarks for chain and many-root topologies: RSS, WASM/native state, disk bytes, append rate, close/reopen time, and local frontier size at 100k and 1M.
[...]

PR #1289, opened 2026-08-16
fix: bound shared-log sync range materialization

peerbit-org:
## Summary

- cap receive-side rateless range materialization at 16,384 entries by default, using a cap-plus-one overflow sentinel
- enforce the cap in TypeScript, shared-log Rust, and native-backbone range paths
- fall back on overflow to the existing exact-hash Simple sync path
- page Simple coordinate lookups and make unsorted in-memory index next(n) lazy and finite

This is the first focused runtime slice following the evidence in #1288 and the investigation in #1286. It does not change wire formats, placement, pruning, or ownership semantics.

## Safety

A rateless overflow sends RequestAll. The source correlates both sync id and peer, then runs Simple sync over the exact outgoing hashes. The fallback does not synthesize a delivery, convergence, or prune acknowledgement.

The in-memory unsorted iterator retains at most one lookahead beyond the requested page and stops raw traversal at the map size captured on first consumption. Sorted iteration retains its existing eager snapshot behavior. Unsorted iteration has documented weak-snapshot behavior under concurrent mutation, and close cancels in-flight next and pending work.

## Architecture follow-up

The audit found that the planner can be entry-free, but execution must page the local frontier with a durable collision-safe keyset cursor. A durable handoff still needs view-fenced plan and move ids, destination retention pins and receipts after all ledger stores cross a durability barrier, concurrent-write catch-up, source prune intents, and restartable idempotent stages. A versioned placement unit is also required: gid is indivisible in the hash domain, while the time domain places entries independently.

## Non-goals

- no durable ownership-transfer protocol in this PR
- no weakening of checked-prune safeguards
- no claim that startup hydration, repair, rebalance, and prune scans are all bounded yet
- no solution for a 100M-entry single gid in the current hash domain

## Verification

- pnpm run build
- pnpm run test
- focused shared-log rateless and Simple-sync tests
- shared-log-rust and native-backbone cargo tests
- indexer-simple Node, browser, and web-worker suites
- ESLint, cargo fmt, and git diff checks

Refs #1286.
Evidence: #1288.

### DIFF
--- a/.changeset/bounded-rateless-ranges.md
+++ b/.changeset/bounded-rateless-ranges.md
@@ -1,0 +1,9 @@
+---
+"@peerbit/shared-log-rust": patch
+"@peerbit/native-backbone": patch
+"@peerbit/shared-log": patch
+---
+
+Bound receive-side rateless range materialization in JavaScript and both native
+backends, fall back to simple sync on overflow, and page simple coordinate
+lookups without using `iterator.all()`.
--- a/.changeset/lazy-simple-index-pages.md
+++ b/.changeset/lazy-simple-index-pages.md
@@ -1,0 +1,8 @@
+---
+"@peerbit/indexer-simple": patch
+---
+
+Page unsorted in-memory index iteration from a bounded forward cursor instead
+of materializing every query match on the first `next()` call. Sorted iteration
+keeps its eager snapshot behavior; unsorted iteration has finite weak-snapshot
+semantics when rows mutate after consumption starts.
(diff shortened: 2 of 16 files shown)


peerbit-org:
Superseded by #1309. The replacement preserves this PR's intent while rebasing onto current master and incorporates the full review: distinct optional limited native capabilities, legacy custom-synchronizer compatibility, snapshot-by-default index behavior, exact cap-plus-one probing, honest overflow benchmark naming/allocation measurement, and refreshed tests/changesets. This branch and its history are left intact; no commits from #1289 were merged.

PR #1290, opened 2026-08-16
refactor: page shared-log rebalance query tasks

peerbit-org:
This is the focused planner slice of the shared-log scaling work. It is independent of the census/evidence PR #1288 and the receive-side sync cap in #1289.

## What changes

- extracts pure rebalance scan planning from index execution
- represents canonical query geometry as plain coordinate DTOs rather than arbitrary range-event objects
- caps each geometry query task at 128 ranges and each task page at 8 total tasks, including boundary work
- keeps interval ownership global across pages so multi-coordinate entries are not duplicated without an emitted-hash set
- preserves observed event order for history mutations and commits them only after every iterator closes successfully
- keeps the existing `toRebalance` execution order and compatibility behavior

## Deliberate scope boundary

This bounds control-plane query construction only. The complete geometry plan is still O(range metadata), and `toRebalance` still drains every matching local row. The numeric task offset is explicitly ephemeral; it is not a restart cursor and cannot authorize pruning.

This PR does not change wire formats, placement semantics, leader selection, transfer acknowledgements, or prune authorization. The next slice will bind progress to exact plan/view identifiers in a bounded alternating two-slot durable work record and use collision-safe frozen hash-number buckets.

## Verification

- focused planner suite: 20 passing for u32/u64
- existing `toRebalance` matrix: 444 passing
- full shared-log node suite: 2,489 passing, 10 pending
- workspace `pnpm run build`: passing
- two independent read-only reviews found no correctness blocker

Refs #1286, #1288, #1289.


peerbit-org:

PR #1291, opened 2026-08-16
feat(shared-log): persist bounded rebalance work

peerbit-org:
## Summary

Stacked on #1290. Review this PR as the single commit after `577b5b38c`.

Adds an internal, bounded restart-state store for immutable rebalance scan plans and collision-safe executor cursors. This is the persistence foundation for the later bounded executor; it is not wired into SharedLog maintenance yet.

## What changes

- persists canonical `{resolution, viewId, plan}` geometry plus a monotone task/collision-bucket cursor
- uses two fixed, alternating checksummed files rather than an append-only WAL
- writes a durable cleared baseline before the first active generation
- requires a named physical barrier in strict mode and re-barriers the selected generation on reopen
- caps frame bytes, geometry, ownership intervals, history mutations, identifiers, and collision buckets before materialization
- treats one hash-number collision bucket as the indivisible restart unit, including a valid empty bucket
- rejects cursor rewinds, task skips, bucket replacement, stale revisions, stale local fences, sparse inputs, noncanonical payloads, and durability-mode resurrection
- poisons the live instance after ambiguous write/barrier failure and resumes only through validated reopen
- settles both A/B reads before fallback or failed-open cleanup
- adds the suite to the existing shared-log CI shard and includes a patch changeset

## Safety boundaries

This PR does **not** change runtime behavior, placement, wire messages, replication semantics, or pruning.

`DurableWorkCommit` and `RebalanceWorkFence` prove only that local restart metadata crossed the configured barrier. They are explicitly not placement epochs, destination receipts, concurrent-write coverage proofs, or prune authorization.

Before runtime wiring, a persistent adapter must provide:

- exclusive ownership/locking for the two-file namespace
- bounded reads that reject oversize before allocating the file
- replace/truncate writes
- a barrier covering the exact file bytes and required namespace metadata

Concurrent-write outboxes/fences, destination retention pins, durable transfer receipts, epoch revalidation, and idempotent staged pruning remain follow-up protocol work.

## Verification

- `pnpm run build`
- `pnpm run test:ci:part-4` — 406 passing
- focused `rebalance work store` suite — 25 passing
- package TypeScript build
- ESLint on the new source/spec
- shard coverage guard
- two independent adversarial reviews of codec, recovery, lifecycle, and boundedness


peerbit-org:

PR #1292, opened 2026-08-17
feat(shared-log): execute bounded rebalance scans

peerbit-org:
## Summary

- add an internal, deterministic rebalance candidate executor over the durable plan/collision-bucket cursor from #1291
- perform one bounded, replayable cursor transition per tick with hard source/visitor row and byte ceilings
- require an atomic complete-bucket source contract, correlated across u32/u64 domains, without using unbounded generic `Index` iteration
- provide stable idempotence keys plus view, abort, deadline, restart, and strict durability fencing
- cover collision filtering, empty buckets, supersession, source/visitor failures, crash replay, ambiguous barrier recovery, and a deterministic small-state model

## Safety boundary

This remains deliberately unwired from `SharedLog` maintenance. It does not transfer entries, apply rebalance history, prune data, change protocol semantics, or treat a durable work-store commit as an ownership receipt. Completion proves only quiescent local candidate enumeration; concurrent-write coverage still requires the planned snapshot fence/outbox. A production source adapter must enforce the supplied limits before allocation; the current generic `Index` interface does not satisfy that contract.

A source without a task-aware secondary index may safely return bounded physical buckets, but that fallback performs one local pass per scan task and is not the eventual 100M-entry backend. The follow-up native source needs a bounded keyset bucket API or composite task/hash index.

## Stack

- based on #1291 (durable bounded work store)
- #1291 is based on #1290 (entry-free deterministic planner)
- separate from evidence-only #1288 and runtime sync-cap #1289

The repository changeset-coverage script currently accepts only PRs targeting `master`, so both the standalone changeset check and Build Workspace's embedded coverage step are expected to remain red while this PR is stacked. They should pass after the predecessor merges and this PR is rebased/retargeted.

## Verification

- `pnpm run build`
- `pnpm run test:ci:part-4` — 425 passing
- focused executor suite — 17 passing
- shared-log TypeScript build
- ESLint on the new source/spec
- CI shard coverage guard


peerbit-org:

PR #1293, opened 2026-08-17
feat(native): read bounded rebalance collision buckets

peerbit-org:
## Summary

- add an authoritative native strict-successor read over the resident shared-log hash-number index
- return either EOF or one complete collision bucket, never a truncated prefix
- preflight caller limits and immutable native ceilings before candidate JS allocation
- preserve u64 keys and coordinates exactly, with typed overflow and fail-closed index-consistency errors
- expose equivalent standalone shared-log Rust and native-backbone wrappers with conformance tests

## Safety boundary

This PR is a plan-agnostic resident-state primitive only. It does not wire the rebalance executor into ordinary maintenance and does not change ownership, transfer, acknowledgement, or pruning behavior.

It intentionally provides no snapshot/write-journal fence, destination receipt, retention pin, or prune authority. Inserts behind a durable cursor and the existing full native-state hydration remain follow-up work.

## Stack

- base: #1292 bounded durable scan executor
- below that: #1291 durable work store and #1290 entry-free scan planner
- #1288 remains the measurement/evidence PR
- #1289 remains the separate receive-side sync-bound slice

Because this is a stacked PR whose base is not `master`, the repository changeset workflow may stop at its base-branch guard until the stack is retargeted.

## Verification

- `pnpm run build`
- shared-log Rust: 28 tests passed
- native-backbone Rust: 58 tests passed
- standalone wasm wrapper: 4 focused tests passed
- native-backbone wasm wrapper: 1 focused test passed
- both package builds and TypeScript no-emit checks passed
- ESLint, Cargo fmt checks, and `git diff --check` passed

### DIFF
--- a/.changeset/native-rebalance-collision-buckets.md
+++ b/.changeset/native-rebalance-collision-buckets.md
@@ -1,0 +1,6 @@
+---
+"@peerbit/shared-log-rust": patch
+"@peerbit/native-backbone": patch
+---
+
+Add bounded, complete native collision-bucket reads for shared-log rebalance scans.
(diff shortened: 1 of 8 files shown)


peerbit-org:

PR #1294, opened 2026-08-17
feat(shared-log): bound rebalance work persistence

peerbit-org:
## Summary

- add optional byte caps to native named-file reads, with pre-allocation checks for memory, Node, OPFS, and buffered adapters
- add a crash-safe `RebalanceWorkStore.beginDrop()` terminal fence that durably clears active restart work before later lifecycle destruction
- cover torn writes, ambiguous barriers, close/drop races, stale instances, exact caps, file growth/shrink, and buffered concurrency
- add patch changesets for `@peerbit/native-backbone` and `@peerbit/shared-log`

## Safety boundary

This is persistence substrate only. It does not wire the scan executor into `SharedLog`, validate a live placement epoch, schedule transfers, acknowledge destination durability, or authorize pruning. `beginDrop()` is local restart metadata, not a transfer receipt or prune capability.

The new optional read caps also do not change ordinary coordinate hydration, whose existing callers still use unbounded reads. Production work-state wiring still requires a lifetime namespace lease, authenticated/current view validation, and a browser lifetime-lock design before durable OPFS use.

## Verification

- `pnpm run build`
- shared-log rebalance work-store suite: 33 passing
- native-backbone Node suite: 122 passing
- native-backbone Rust suite: 58 passing
- targeted TypeScript, lint, Prettier, and `git diff --check`
- independent final adversarial review: no blocker

## Stack

This PR is based on #1293, which is based on #1292, #1291, and #1290. Evidence PR #1288 and sync-cap PR #1289 remain separate. The stacked changeset base guard may remain red until the PR is retargeted after its base merges.

### DIFF
--- a/.changeset/bounded-native-named-file-reads.md
+++ b/.changeset/bounded-native-named-file-reads.md
@@ -1,0 +1,6 @@
+---
+"@peerbit/native-backbone": patch
+"@peerbit/shared-log": patch
+---
+
+Add pre-allocation byte caps to native named-file reads and a crash-safe terminal fence for durable rebalance work.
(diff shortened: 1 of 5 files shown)


peerbit-org:

PR #1295, opened 2026-08-17
feat(shared-log): lease rebalance work persistence

peerbit-org:
## Summary

- add an internal bounded memory adapter for non-persistent work-state restart tests
- add an atomic Node factory that opens `RebalanceWorkStore` together with a canonical per-log lifetime lease
- durably create and parent-fsync `rebalance-work/<hex-log-id>`, then strictly sync the exact namespace after every raw named-file barrier
- serialize all reads, writes, barriers, flushes, and close under the OS-released lease
- reject symlink namespace crossing, limit widening, oversized reads, stale ownership, and cleanup ambiguity
- add direct real-filesystem and injected-failure coverage plus a shared-log patch changeset

## Safety boundary

This module is internal and is not exported from the package root or wired into live `SharedLog`. It only supplies opaque bounded persistence. It does not validate placement-view currency, schedule scans, transfer entries, acknowledge destination durability, create retention pins, or authorize pruning.

The Node factory returns an already-open work store rather than exposing its leased persistence object, so setup, preflight, and recovery failures release raw resources and the namespace lease atomically. Browser durable persistence remains unsupported until a worker/lifetime-lock design exists.

## Verification

- `pnpm run build`
- combined rebalance work-store shard: 46 passing
- new persistence suite: 13 passing
- shared-log TypeScript and targeted ESLint
- Prettier, CI shard-coverage guard, and `git diff --check`
- two independent adversarial reviews: no blocker

## Stack

This PR is based on #1294, then #1293, #1292, #1291, and #1290. Evidence PR #1288 and sync-cap PR #1289 remain separate. The stacked base/changeset guard may remain red until the PR is retargeted after its base merges.

### DIFF
--- a/.changeset/leased-rebalance-work-persistence.md
+++ b/.changeset/leased-rebalance-work-persistence.md
@@ -1,0 +1,5 @@
+---
+"@peerbit/shared-log": patch
+---
+
+Add an internal bounded memory adapter and an atomically leased Node persistence factory for durable rebalance work.
--- a/packages/programs/data/shared-log/src/rebalance-work-persistence.ts
+++ b/packages/programs/data/shared-log/src/rebalance-work-persistence.ts
@@ -1,0 +1,531 @@
+import { toHexString } from "@peerbit/crypto";
+import type {
+	NativeBackboneCoordinatePersistenceStore,
+	NativeDurabilityLease,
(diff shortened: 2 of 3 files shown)


peerbit-org:

PR #1296, opened 2026-08-17
feat(shared-log): fingerprint local placement views

peerbit-org:
## Summary

- add a bounded canonical binary V1 fingerprint for complete frozen local placement facts
- bind explicit planner, domain, policy, identity, range, maturity, and tri-state peer-filter inputs under hard collection and byte caps
- keep the semantic view digest separate from a per-open execution fence, with pinned binary vectors and adversarial validation tests

Stacked on #1295.

## Safety boundary

This is a receiver-local placement-view fingerprint, not consensus, a global placement epoch, a custody receipt, or pruning authority.

This PR does not wire runtime capture, owner planning, transfer, acknowledgements, or deletion. Future execution must still use bounded paged capture of freshly authenticated facts, stable ownership/role guards, a durable destination custody acknowledgement, and final live placement revalidation.

## Tests

- `pnpm exec prettier --check packages/programs/data/shared-log/src/local-placement-view.ts packages/programs/data/shared-log/test/local-placement-view.spec.ts .changeset/local-placement-view-fence.md package.json`
- `pnpm --filter @peerbit/shared-log exec tsc --noEmit --project tsconfig.json`
- `pnpm --filter @peerbit/shared-log exec eslint src/local-placement-view.ts test/local-placement-view.spec.ts`
- `node scripts/ci/check-shard-coverage.mjs`
- `node ./node_modules/aegir/src/index.js run test --roots ./packages/programs/data/shared-log -- -t node --grep "^local placement view"`
- `git diff --check`

PR #1297, opened 2026-08-17
fix(shared-log): make owner sampling deterministic

peerbit-org:
## Summary

- count mature placement evidence per owner in both TypeScript and Rust, regardless of which overlapping range is observed first
- keep Rust's optimized membership check aligned with full leader selection
- make TypeScript closest-range ties use the native total order: distance, timestamp, owner hash, then range ID
- fix Rust descending coordinate searches so equal-coordinate candidates keep metadata ascending, including wrapped fallback
- add direct JS, Rust-core, and u32/u64 JS-to-WASM parity regressions

Stacked on #1296. The repository's master-only changeset/build guards remain red for stacked PR bases and will become actionable when this slice is retargeted after its parent lands.

## Runtime compatibility

This intentionally corrects placement behavior in mixed-maturity and exact-tie corner cases where JavaScript and Rust could already disagree. The shared closest-iterator tie rule also affects related closest/cover/adjacency consumers.

Mixed old/new deployments can therefore disagree on these corner cases. The future bounded planner must bind an explicit planner semantic version in the placement-view digest.

## Scope boundary

This is a semantics/parity fix, not the bounded planner. It does not remove existing Index `.all()` paths, the native fallback's quadratic worst case, containment-bucket amplification, or add transfer/receipt/prune behavior.

## Tests

- Rust core: 31/31
- shared-log ranges: 1504/1504
- native planner/parity: 132/132
- focused u32/u64 differential regressions: 6/6
- TypeScript builds for both packages
- ESLint and Prettier
- `cargo fmt --check`
- shard coverage guard
- `git diff --check`

### DIFF
--- a/.changeset/young-owners-agree.md
+++ b/.changeset/young-owners-agree.md
@@ -1,0 +1,7 @@
+---
+"@peerbit/shared-log": patch
+"@peerbit/shared-log-rust": patch
+---
+
+Make shared-log owner sampling independent of mixed-maturity range order and
+align TypeScript fallback tie-breaking with the native deterministic order.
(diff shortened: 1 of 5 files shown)


peerbit-org:

PR #1298, opened 2026-08-17
perf(shared-log): bound native containment indexing

peerbit-org:
## Summary

- cap native RangePlanner containment fanout at eight fine buckets per narrow range
- store wider ranges once in an ordered spill set
- merge fine and spill candidates in exact put/re-put order
- release empty and disproportionate retained IndexSet capacity under churn

## Motivation

The previous 4096-bucket containment index cloned every range ID into every covered bucket. At the frozen-view ceiling used by the scaling work, 16,384 full-width ranges retained 67,108,864 containment entries before ordinary planner state. This makes a bounded placement view insufficient to bound native planner memory.

This patch changes only the native in-memory containment index representation. It preserves owner membership, cursor-major ordering, maturity behavior, fallback behavior, and wire/protocol semantics.

## Bound

- narrow range: at most 8 fine-bucket IDs plus 1 global-order ID
- wide range: 1 spill ID plus 1 global-order ID
- retained capacity: compacted when a set exceeds 2 x live entries + 8 slots; empty sets and clear release their allocations

The wide spill still makes a containment lookup O(number of wide ranges), and global ordered removal is still O(number of ranges). Those execution-cost bounds remain follow-up work; this PR addresses physical index amplification only.

## Tests

- 16,384 full-width ranges retain 16,384 spill IDs rather than 67,108,864 bucket IDs
- exact 8-bucket versus 9-bucket threshold, wrapped geometry, u32/u64 maximum boundaries
- interleaved narrow/wide order, re-put, delete, narrow-to-wide, wide-to-narrow, zero-width, clear
- bucket migration and spill contraction release retained capacity

Verification:

- shared-log Rust: 38/38
- native-backbone Rust: 58/58
- rebuilt native/TypeScript planner parity: 132/132
- all three package builds
- cargo fmt, Changesets guard 52/52, Prettier, diff check

## Stack

Stacked on #1297 because this representation relies on the deterministic owner-sampling semantics fixed there. Master-only/merge-base guard jobs are expected to remain red while the PR targets a stack branch; they become actionable when the stack is retargeted to master.

### DIFF
--- a/.changeset/bounded-containment-index.md
+++ b/.changeset/bounded-containment-index.md
@@ -1,0 +1,8 @@
+---
+"@peerbit/shared-log": patch
+"@peerbit/shared-log-rust": patch
+"@peerbit/native-backbone": patch
+---
+
+Bound native shared-log containment indexing for wide replication ranges while
+preserving exact owner-sampling membership and ordering.
--- a/packages/programs/data/shared-log/rust/src/lib.rs
+++ b/packages/programs/data/shared-log/rust/src/lib.rs
@@ -4,6 +4,7 @@ use sha2::{Digest, Sha256};
 use std::cmp::Ordering;
 use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::ops::Bound::{Excluded, Unbounded};
+use std::ops::ControlFlow;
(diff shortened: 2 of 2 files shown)


peerbit-org:

PR #1299, opened 2026-08-17
perf(shared-log): bound native fallback planning

peerbit-org:
Stacked on #1298. Review only commit `e44dc425e` for this slice.

## What this changes

- Replaces the native fallback planner repeated restart-and-skip search with four stateful cyclic endpoint streams.
- Bounds native fallback endpoint advances to at most four passes over the fallback index per cursor, changing the adversarial path from quadratic to linear in the indexed range facts.
- Stores endpoint metadata in coordinate-keyed ordered sets, preserving deterministic distance, timestamp, owner, and range-id ordering without materializing tie groups.
- Removes the synthetic maximum-string sentinel and correctly merges maximum-timestamp/high-Unicode identifiers and circularly aliased zero/MAX endpoints.
- Makes the TypeScript fallback phases disjoint and monotone, treats exact directional equality as zero distance, and excludes the current full-width range from adjacency lookup.
- Fixes the joined iterator lifecycle so prefetched rows count as pending, prevent premature done, remain ordered through `all()`, and are released on close.

## Evidence

- shared-log Rust: 42 passing
- TypeScript/native parity: 136 passing, including seeded u32/u64 sweeps
- full shared-log ranges suite: 1,518 passing
- native-backbone Rust: 58 passing
- changeset guard: 52 passing
- package builds, changed-file ESLint, Rust formatting, Prettier, and diff checks pass

The 16,384-range adversarial test records exactly four endpoint-index advances per range for one cursor. The prefetched-tail regression uses 150 disjoint fallback rows and proves a mature owner beyond batch 100 is not truncated.

## Compatibility note

This intentionally corrects rare fallback ordering at exact endpoints, zero/MAX aliases, and maximum-timestamp/high-Unicode metadata. Mixed-version peers can therefore choose different fallback owners for those boundary/tie cases until the stack is upgraded. Ordinary intersecting ownership is unchanged.

## Scope boundary

This does not make the generic TypeScript Index backend scan-bounded, remove its existing `.all()` materialization paths, bound startup hydration, or add transfer receipts, custody pins, concurrent-write capture, or prune authority. It is a prerequisite for a deterministic bounded owner planner over a capped frozen placement view.

Because this PR is stacked on #1298, master-only ancestry and changeset checks are expected to remain red until the stack is rebased or merged.

### DIFF
--- a/.changeset/linear-native-fallback-streams.md
+++ b/.changeset/linear-native-fallback-streams.md
@@ -1,0 +1,16 @@
+---
+"@peerbit/shared-log": patch
+"@peerbit/shared-log-rust": patch
+"@peerbit/native-backbone": patch
+---
+
+Bound native shared-log fallback sampling to linear endpoint-index work per
+cursor while preserving deterministic owner selection. Replace the synthetic
+maximum metadata sentinel with coordinate-only partitions so maximum-timestamp,
+high-Unicode identifiers remain in their correct fallback segment, and apply
+metadata tie-breaking across circularly aliased zero/MAX endpoints. Split the
+TypeScript directional fallback into disjoint monotone phases, treat exact
+directional equality as zero distance, and exclude the current range from
+same-owner adjacency queries. Keep prefetched join rows visible to iterator
+completion and pending counts so batched fallback scans cannot truncate them,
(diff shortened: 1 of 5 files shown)


peerbit-org:

PR #1300, opened 2026-08-17
feat(shared-log): plan owners from frozen placement views

peerbit-org:
## Summary

Stacked on #1299.

This adds the next non-destructive runtime primitive for the shared-log scaling work: compile one complete, capped, locally authenticated placement view into an isolated native owner planner, then answer deterministic bounded owner queries without reading an Index.

- accept only canonical placement views issued in the current process and revalidate their bytes/digest before native allocation
- hydrate the complete frozen range set once, converting canonical range IDs back to the runtime base64 representation
- expose bounded u32/u64 plans with exact dense coordinates, the protocol's 100-replica ceiling, canonical owner ordering, immutable authenticated key rows, and a versioned `planId`
- pass every captured policy/time/filter/self-role input explicitly; planning never uses `Date.now()` or live runtime state
- give standalone native range planners an explicit idempotent `close()`/`free()` lifecycle and reject work before consuming inputs after release
- close partial hydration exactly once and preserve cleanup failures

## Boundedness

For an issued view, construction is bounded by the existing view ceilings (4,096 owners / 16,384 ranges / 4 MiB). Each plan accepts at most 100 coordinates, validates native output against the issued owner set before materializing rows, and caps canonical plan-identity encoding at 4 MiB.

This module has no Index dependency and no `.all()` path.

## Safety boundary

The view digest and `planId` are local deterministic decision identities only. They are not a global placement epoch, transfer acknowledgement, custody receipt, retention pin, or pruning authority. This PR does not schedule runtime work, transfer entries, or delete data.

Before destructive maintenance is possible, follow-ups still need bounded fresh-view capture with current V2 provenance, a concurrent-write fence/outbox, destination durable custody receipts and pins, a source receipt/prune journal, and a final lane-fenced placement revalidation.

## Tests

- local placement view + owner planner: 25 passing
- pure owner planner: 12 passing (real u32/u64 WASM)
- Rust core: 42 passing
- native planner close/lifecycle: passing
- shared-log and shared-log-rust builds
- changed-file ESLint
- shard coverage guard
- `git diff --check`

### DIFF
--- a/.changeset/pure-local-owner-plans.md
+++ b/.changeset/pure-local-owner-plans.md
@@ -1,0 +1,9 @@
+---
+"@peerbit/shared-log": patch
+"@peerbit/shared-log-rust": patch
+---
+
+Add an internal, issued-view-only owner planner that hydrates the native range
+planner once and returns canonical bounded decisions without querying an Index.
+Give standalone native range planners an explicit idempotent close/free
+lifecycle and reject every operation after release.
--- a/packages/programs/data/shared-log/rust/src/index.ts
+++ b/packages/programs/data/shared-log/rust/src/index.ts
@@ -227,6 +227,7 @@ type ResidentRepairDispatchPlanInput = Omit<
 type RepairDispatchPlan = Map<string, Map<string, string[]>>;
 
 type NativeRangePlannerHandle = {
+	free: () => void;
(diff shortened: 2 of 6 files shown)


peerbit-org:

PR #1301, opened 2026-08-17
feat(shared-log): snapshot native placement ranges

peerbit-org:
## Summary

- add complete-or-overflow native snapshots of resident replication-range geometry and owner hashes
- enforce fixed owner, range, per-owner, identifier, and logical-byte ceilings before exposing any row to JavaScript
- validate deterministic UTF-8 ordering, exact u32/u64 geometry/accounting, and secondary owner-index consistency
- expose the same bounded primitive through standalone range planners, shared native state, and native-backbone

## Safety boundary

This is an unwired resident-state primitive only. It does not query an `Index`, authenticate owner hashes, create a placement epoch, transfer entries, issue custody receipts, or authorize pruning. Runtime capture still requires current V2 owner provenance and the global ownership-mutation lane. Native hydration and concurrent-write/outbox coverage remain separate follow-up work.

Stacked on #1300.

## Verification

- shared-log Rust: 49/49
- native-backbone Rust: 58/58
- shared-log Node/WASM: 143/143
- native-backbone Node/WASM: 123/123
- both package builds and lints
- `cargo fmt --check`
- changeset guard tests: 52/52
- `git diff --check`
- two independent safety reviews: no blockers

### DIFF
--- a/.changeset/bounded-native-range-snapshots.md
+++ b/.changeset/bounded-native-range-snapshots.md
@@ -1,0 +1,7 @@
+---
+"@peerbit/shared-log-rust": patch
+"@peerbit/native-backbone": patch
+---
+
+Add complete-or-overflow bounded snapshots of resident native shared-log range
+geometry and unauthenticated owner hashes, including native-backbone forwarding.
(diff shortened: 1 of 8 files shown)


peerbit-org:

PR #1302, opened 2026-08-17
feat(shared-log): fence current V2 owner provenance

peerbit-org:
## Stack

Stacked on #1301. Review the top commit f3cc58741 for this slice.

## Summary

- track one bounded volatile current-owner token per authenticated replication-info V2 peer, bound to the exact open generation, peer/receive sessions, transport sessions, challenge/binding, sender epoch, sequence, and receiver-local monotonic grant
- publish provenance only after Full, Added, or Stopped ownership work durably commits and all fallible bookkeeping finishes inside the global ownership mutation lane
- invalidate tokens and branded all-or-nothing captures across mutation start, failure, expiry, clock regression, lifecycle rotation, session/transport change, recovery, close, and reopen
- move unknown and refreshed-empty Stopped completion into the same ownership lane, and fail closed when current authorization rejects or publication cannot complete

## Safety boundary

This slice has no placement snapshot consumer, executor wiring, transfer handler, custody receipt, or prune path. The token is volatile current-session evidence only; it is explicitly not consensus, custody, transfer, or prune authority, and future capture must re-evaluate custom replicator authorization.

## Verification

- complete shared-log Node suite: 2,604 passing, 10 pending
- final focused receiver state and integration suite: 45 passing
- shared-log TypeScript build and no-emit typecheck
- shard coverage guard
- changeset guard tests: 52 passing
- git diff check
- three independent read-only safety reviews: no blockers

### DIFF
--- a/.changeset/current-v2-owner-provenance.md
+++ b/.changeset/current-v2-owner-provenance.md
@@ -1,0 +1,5 @@
+---
+"@peerbit/shared-log": patch
+---
+
+Track bounded, current-session V2 remote-owner provenance, invalidate it across lifecycle, session, transport, recovery, and clock fences, and publish it only after ownership mutations fully finalize inside the global mutation lane. This volatile evidence is for future placement capture only and carries no custody or prune authority.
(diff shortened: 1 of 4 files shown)


peerbit-org:

PR #1303, opened 2026-08-17
feat(shared-log): model bounded custody handoffs

peerbit-org:
## Summary

- add a canonical V1 codec for one-entry, one-destination custody handoffs and signed receipts, bound to an exact rebalance visit and fixed custody profile
- add an internal lazy point-record model with bounded admission, monotone source/destination reducers, logical A/B generations, strict barriers, fail-closed recovery, and ambiguous-write poisoning
- add golden, adversarial, crash-cut, lifecycle, and resource-bound coverage and wire both suites into the shared-log CI shard

## Safety boundary

This PR is deliberately internal and unwired. It adds no runtime scheduler, network handler, transfer finalization, enumeration, release, deletion, or prune authority, and neither module is root-exported. Memory mode is only a reducer simulation: it withholds receipt bytes and never returns durable commit evidence.

Production use remains blocked on a real leased keyed CAS backend, exact log/local-key/role namespace binding, a lower-store composite durability-and-pin issuer, durable move discovery and entry-to-pin lookup, a bounded release lifecycle, and current placement/custody revalidation at the final delete boundary. Recovered receipts and pin facts are historical evidence only.

## Validation

- focused custody codec/store suites: 28 passing
- shared-log build and TypeScript check
- shared-log shard coverage guard
- changeset guard: 52 passing
- git diff check

Stacked on #1302.


peerbit-org:

PR #1304, opened 2026-08-17
feat(shared-log): persist keyed custody records

peerbit-org:
## Summary

- add an internal Node-only custody-record persistence factory backed by dedicated SQLite point rows, cap-plus-one BLOB reads, strict WAL checkpoints, and namespace-directory fsync
- bind every database and record generation to the exact node root, log, canonical local public key, source/destination role, custody profile, durable domain, writer epoch, and writer owner
- add a crash-released fence-less directory lock that shares the existing native durability lock namespace without reading or rewriting the persisted fence record
- make the custody reducer validate namespace binding on every recovered and newly admitted artifact before it can return a strict durable commit

## Safety boundary

This remains an internal, unwired persistence prerequisite. It does **not** schedule or transfer entries, mint destination pin evidence, release custody, delete data, or authorize pruning. The database `namespaceEpoch` is storage identity only and is explicitly distinct from the protocol receipt `custodyEpoch`. Browser and OPFS persistence remain unsupported.

The next stacked slice will add a single-flight durable `pendingVisit` bridge so an in-flight handoff stays discoverable across work-plan replacement and restart. Production pin enforcement, release/GC, concurrent-write outbox handling, and prune authorization remain later work.

## Verification

- shared-log custody codec/persistence/store: 41 passing
- native-backbone Rust: 58 passing
- native directory lock: 10 passing
- shared-log and native-backbone TypeScript/WASM builds
- shard coverage guard
- changeset guard: 52 passing
- `git diff --check`

Stacked on #1303.


peerbit-org:

PR #1305, opened 2026-08-17
feat(shared-log): persist custody visits before effects

peerbit-org:
## Summary

- persist one exact signed custody manifest before any custody or network callback can run
- mirror pending state through both strict A/B barriers, repair it on reopen, and publish it in memory only after both barriers succeed
- advance the cursor only from an exact paired source-receipt durability proof or an explicit side-effect-free bounded point-read skip
- allow old-view pending copies to finish behind a separate namespace/lifecycle recovery guard while new work remains placement-guarded
- bind the bridge to the exact strict source custody generation, log ID, source key, work plan, task, bucket, entry, and handoff generation
- harden caller and persistence byte captures with typed-array intrinsics before allocation

## Safety boundary

This remains an internal, unwired bridge. It adds no release, transfer-finalization, ownership, deletion, or prune authority. The preparer/skip point read, visitor idempotence, and recovery lifecycle guard are still trusted integration seams. Raw CustodyRecordStore transitions do not establish the WorkStore ordering invariant by themselves.

## Verification

- shared-log CI part 4: 526 passing, discovery assertion passed
- focused custody/work/executor/model suite: 113 passing
- changeset guard: 52 passing
- shared-log build
- changed-file ESLint
- Prettier
- git diff --check

## Stack

This PR is stacked on #1304, which adds the leased keyed custody persistence adapter.


peerbit-org:

PR #1373, opened 2026-09-01
feat: add crash-safe two-slot checkpoints

### LINKED ISSUE #1286: Scale shared-log to 100M global entries with bounded per-peer state

peerbit-org:
## Summary

- add an optional crash-safe atomic replacement capability to the AnyStore interface
- implement it only for ClassicLevel-backed root stores and sublevels with one atomic batch and sync fencing
- capture exact Uint8Array bytes before returning control and reject proxies, detached buffers, or other typed arrays
- add an opt-in @peerbit/any-store/checkpoint subpath with two fixed alternating slots, strict checksums, scope binding, generation parity, and predecessor links
- reject every malformed present slot, ambiguous replacement, concurrent commit, unsupported backend, overflow, and over-bound payload

## Crash and resource model

The helper is single-writer, requires a trusted dedicated store/sublevel, accepts at most 64 MiB per payload, performs sequential slot reads, and keeps only two logical keys. A failed replacement makes the instance terminal until reopen.

Real ClassicLevel tests prove an acknowledged generation-two checkpoint survives SIGKILL and reopens exactly. Fault injection covers old-or-new replacement, applied-then-error, torn state, corrupt or mixed slots, and terminal ambiguity.

The helper does not claim compare-and-swap, multi-process coordination, writer authentication, hostile-backend bounded reads, physical LSM compaction, whole-snapshot rollback detection, log truncation, or tombstone compaction. Memory, OPFS, Rust, proxy, and no-op stores remain capability-absent and fail closed.

## Compatibility and bloat

This is additive and opt-in. atomicReplace is optional, the checkpoint is not re-exported from the package root, and existing root imports/load graphs are unchanged. Patch changesets match the directly comparable public durability-capability precedent in f19a1c0cb.

Against exact master, the tarball adds about 12.8 KiB compressed / 50.4 KiB unpacked and 11 entries. Tests are excluded. Direct production dependencies fall from 10 to 7 and unique production name/version nodes fall from 83 to 82; unused crypto/logger runtime edges are removed and the checkpoint uses the already-present small @stablelib/sha256 package.

## Validation

- AnyStore matrix: 62 Node, 58 browser, 58 web-worker tests
- real ClassicLevel root/sublevel durability and SIGKILL/reopen tests
- full monorepo build
- frozen-lockfile install and dependency checks
- root lint: zero errors, 15 pre-existing warnings
- changeset, formatting, diff, pack, and subpath-only import checks
- independent review after fixes: no remaining P0–P3 findings

This enables a later replacement of TrustedNetwork's internal append anchor and a measured bounded-state slice. It does not by itself truncate shared-log history or close #1286.


peerbit-org:
Local diagnostic on exact head f6c265880c4b4af1686eb10988bdb031fbaed3d0 found no performance/storage blocker.

Matched 100k x 512-byte durable commits on Apple M3 Pro / APFS / Node 24.13.1 / classic-level 3.0.0:

| Metric | Two-slot checkpoint | Current append + barrier model |
|---|---:|---:|
| Median write time | 5.119 s | 10.967 s |
| Throughput | 19,534 ops/s | 9,118 ops/s |
| Restore only | 0.279 ms | 424.667 ms |
| Level open + restore | 6.112 ms | 431.681 ms |
| Logical state | 2 keys / 1,264 B | 100,000 keys / 63.2 MB |
| Directory after reopen | 3.715 MB | 37.322 MB |
| End RSS after reopen | 85.0 MB | 228.4 MB |

Raw 100k write samples:

- checkpoint: 5.119, 6.517, 4.770 s
- append: 9.494, 10.967, 12.331 s

Important nuance: the current append model performs a sync put plus a second sync barrier. A one-sync append counterfactual reached 4.096 s / 24,417 ops/s, about 25% faster writes than A/B. The checkpoint's decisive gains are bounded logical state, reopen, and retained storage; the 2.14x write gain mostly removes the current duplicate physical sync.

Physical LevelDB state behaved well but is not a protocol bound: the checkpoint directory was about 4.83 MB at 10k and 4.90 MB at 100k. Manual compaction took 2.25 ms and reduced the reopened directory from 3.715 MB to 10.3 KB. The README caveat remains necessary.

Worst relevant fork snapshot (66 x 128 KiB = 8,650,752 B): median durable commit 47.65 ms; checkpoint restore 31.49 ms; open + restore 56.75 ms; directory 8.67 MB. Peak RSS was roughly 178 MB in the warmed process. TrustedNetwork integration should therefore configure a tight ~8.25 MiB plus framing bound rather than inherit the generic 64 MiB ceiling.

This is local diagnostic evidence, not a cross-machine performance gate. The benchmark modeled the current immutable-generation existence check, sync put, fresh barrier, predecessor checksum chain, iteration/sort/get restore, and exact payload recovery; it omitted Borsh/authentication/reducer work, making the append reopen comparison conservative.

peerbit-org:
Integration sizing correction from the follow-up TrustedNetwork preflight: the 8.25 MiB benchmark payload modeled the maximum 66 child proofs only. A self-contained FORKED anchor checkpoint must also retain the common-parent entry.

For the proposed fixed Borsh payload, the exact application payload ceiling is 8,782,137 bytes (about 8.3753 MiB):

3-byte variant tags + state byte + (4 + 131,072)-byte head + 4-byte optional candidate tag/length + 32-byte digest + 4-byte reason length + 1-byte empty reason + 66 x (4 + 131,072)-byte proof vectors.

The generic checkpoint record adds 120 bytes, for 8,782,257 bytes total. This remains comfortably below #1373's 64 MiB generic ceiling and does not change the utility PR verdict, but the later TrustedNetwork integration should configure the exact 8,782,137-byte application bound rather than the rough 8.25 MiB figure.

Issue #1346, opened 2026-08-30
TrustedNetwork v2: roles, deterministic revocation, and future-write confidentiality


peerbit-org:
## Problem

`TrustedNetwork` currently has one undifferentiated trusted role. Revocation authorization can be tightened locally, but the larger protocol still lacks:

- separate admin, writer, reader, and replicator authority;
- a deterministic revocation boundary across replicas;
- an explicit historical-data policy after revocation; and
- future-write confidentiality from a removed member.

A `canRead` predicate alone is not confidentiality: a peer that can obtain replicated blocks can retain or redistribute plaintext. Likewise, validating trust only when an entry is first ingested leaves races around revocation and offline replicas unless the protocol defines a common authorization boundary.

PR #1344 is intentionally only the small owner-authorized revoke primitive. This issue tracks the versioned protocol work that should not be folded into that patch.

Related: #51 (authorization re-check/revalidation), #290 (parent key/certificate revocation).

## Required invariants

1. Roles are explicit and least-privilege: admin, writer, reader, and replicator capabilities are independently expressible.
2. Every replica reaches the same authorization verdict for a write at a defined policy version/epoch.
3. Revocation semantics state whether old ciphertext/history remains readable and what happens to writes racing the boundary.
4. Mixed-version peers fail closed when they cannot validate the new policy format.
5. Removed readers cannot decrypt future writes after an epoch transition.
6. State, key distribution, replay work, and network fanout remain bounded.
7. Crash/reopen and offline catch-up cannot restore revoked authority.

### A. Versioned membership records

Introduce an explicitly versioned v2 record rather than changing `IdentityRelation` in place. Include a role bitset, monotonic policy sequence or epoch, subject, issuer, and authorization proof. Old validators must not interpret a v2-controlled store as legacy permissive state.

### B. Deterministic revocation boundary

Choose and specify one convergent model before implementation:

- a linearized, authority-signed policy head with writes bound to the policy sequence they were authorized under; or
- short-lived, authority-signed capabilities with deterministic expiry and renewal rules.

The design must resolve the offline-old-head case and define revalidation/rollback behavior. Merely deleting a trust relation at local arrival time is insufficient.

### C. Epoch-key content encryption

[...]

PR #1376, opened 2026-09-01
feat(trusted-network): use bounded policy checkpoints

peerbit-org:
## Summary

- replace the internal TrustedNetwork v2 append-only policy-anchor history with the scope-bound crash-safe two-slot checkpoint from `@peerbit/any-store/checkpoint`
- publish ACTIVE, UNAVAILABLE, and complete canonical FORKED state with one atomic replacement, while preserving authorization fencing, the bounded admission queue, and terminal fail-stop behavior
- bind records to the exact canonical network descriptor and reject the superseded append namespace rather than adding an unneeded migration/fallback path
- retain the complete bounded 2–66 observation set so peerless reopen reconstructs the same canonical fork evidence
- remove generation keys, predecessor chains, per-generation scans, and duplicate durable fences

## Compatibility

This changes an internal module that is intentionally absent from the package entry point. It does not change the public TrustedNetwork wire format or public exports, and it carries a patch changeset.

The superseded append format was internal and never activated. If those legacy keys are found in the dedicated store, open fails closed and requires an explicit reset; silently treating them as an empty checkpoint would create a rollback path. There is no legacy fallback retained in the package.

The checkpoint remains single-writer and does not claim cross-process CAS. Its logical state is exactly two slots; physical LevelDB files remain subject to normal LSM compaction.

## Validation

- frozen workspace install
- target package build
- full TrustedNetwork suite: 76 passing
- post-rebase focused anchor/hard-kill matrix: 26 passing, including three real SIGKILL recovery boundaries
- semantic valid-envelope corruption, exact 66-observation restore, valid 67th rejection, descriptor mismatch, mixed legacy/checkpoint rejection, ambiguous commit, bounded queue, hostile typed arrays, and zero post-FORK work
- independent review: no remaining P0–P3 findings
- package tarball: 82,024 → 74,632 bytes compressed (-9.01%); 416,914 → 372,468 unpacked (-10.66%)

## Real 100k signed-policy benchmark

Exact merged append anchor `3e683511f` versus this change, using one pre-generated 46,699,947-byte stream of 100,000 valid sequential authority-signed `EntryV0` policies. Fixture generation was excluded. Three fresh processes/directories per cohort were counterbalanced; every run reopened the exact ACTIVE head at sequence 99,999, with the same digest, zero pending work, and zero resolver calls.

| Metric | Append anchor | Two-slot checkpoint | Change |
| --- | ---: | ---: | ---: |
| Admission wall, median | 22.037 s | 16.292 s | -26.07% |
| Admission CPU, median | 20.092 s | 16.046 s | -20.13% |
| Update rate | 4,538/s | 6,138/s | +35.26% |
| Restore only | 1,245.6 ms | 7.8 ms | 160.3x faster |
| Level open + restore | 1,254.2 ms | 17.8 ms | 70.6x faster |
| Logical state | 100,000 keys / 74,599,947 B | 2 keys / 1,330 B | 56,090x smaller |
| Apparent disk before close | 42.55 MB | 5.59 MB | -86.86% |
| Apparent disk after first reopen | 42.68 MB | 4.62 MB | -89.18% |
| Fresh-reader peak RSS | 447.1 MB | 118.4 MB | -73.53% |

An independent fresh 100k pair reproduced the direction: 18.44% lower write wall time and 169.7x faster restore. A maximum 8,782,137-byte application payload plus the 120-byte generic envelope committed in 40 ms median.

Progresses #1346 and the bounded-state work in #1286.


peerbit-org:
### Raw real 100k benchmark samples

Environment: Apple M3 Pro / 36 GiB / macOS 26.6.2 / Node 24.13.1 / ClassicLevel. Exact append cohort `3e683511f`; checkpoint cohort `6af495f` (source/test/doc blobs identical to rebased PR head `b19d4d0a`). One pre-generated 46,699,947-byte signed fixture, SHA-256 `ea7d25876ed96a47229f0d74af1cfde58eaff27952546ffb403750f216b41ab7`; generation excluded. Fresh-process/fresh-directory order: append, checkpoint, checkpoint, append, append, checkpoint.

Admission samples are wall ms / CPU ms / updates per second:

- append: `18722.974 / 19047.146 / 5341.032`, `22037.316 / 20091.535 / 4537.758`, `23567.376 / 22995.950 / 4243.154`
- checkpoint: `15426.625 / 15969.244 / 6482.299`, `17069.656 / 16426.230 / 5858.349`, `16292.198 / 16046.332 / 6137.907`

Restore-only wall samples:

- append: `1151.889`, `1245.593`, `1492.552` ms
- checkpoint: `7.157`, `15.553`, `7.772` ms

Every writer and fresh reader recovered ACTIVE sequence `99999`, digest `b20ac9d32af7fc50b165f398f6337d35f2a86b84de39a34bec13401f1ba9dee6`, pending `0/0`, and resolver calls `0`.

Logical state was stable in every sample: append `100000` keys / `74599947` combined bytes; checkpoint exactly `2` keys / `1330` combined bytes, holding generations `99999` and `100000`.

Physical sizes use natural LevelDB behavior with no forced compaction. They are intentionally reported as observed rather than claimed as a strict file-size bound; only the two-key logical state is strict.

PR #1407, opened 2026-09-02
fix: bound stale-provider cold joins

peerbit-org:
## Context

Investigates the minute-scale `Documents.open` / SharedLog cold-join mode reported in dao-xyz/peerbit-examples#291. The downstream 10-run profile isolated 63–74 second stalls at the document-store open boundary when reachable donors coexisted with recently stopped replicas.

## Summary

- bound initial fanout bootstrap dialing, candidate dialing, join writes/replies, cooldowns, and topic-root query rounds by shared overall deadlines
- launch root queries concurrently while preserving stable peer-order precedence, and cancel stalled sends/losing queries
- cancel timed-out unsent JOIN requests and always clear pending join state so a late write cannot create a one-sided ghost child
- prefer current reachable fanout candidates and stop discovery after the first usable live bootstrap/candidate
- retain responsive bootstrap trackers, exclude stale authenticated redials, and rotate an unhelpful tracker cohort after an unsuccessful round
- keep proactive 256-shard host scans on local resolver/tracker state, cap them at eight workers under one shared discovery deadline, cancel sibling work on resolver failure, and drain any joins already started
- locally suppress only the exact signed auto-root claim of a directly departed origin; preserve its replay floor and reactivate it on exact authenticated reconnect or any newer valid signed claim
- widen refreshed SharedLog provider discovery beyond the first eight already-tried stale candidates while retaining the normal bounded lookup
- expose aggregate SharedLog open/fanout/provider-resolution profile events through the existing optional `sync.profile` callback

## Safety and compatibility

This is additive at the API and format boundaries. There are no wire-format, persisted-format, dependency, or migration changes. Ranked-strict and weighted fanout ordering are preserved; only the default ranked-shuffle path promotes one currently reachable candidate. Tracker feedback is best-effort and cannot hold open completion. Proactive shard hosting no longer queries connected peers per shard; lazy subscribe/open discovery still does.

Direct-origin suppression is deliberately a local failure-detector heuristic, not a revocation or global dead-root proof. If a live origin loses its one direct edge to a node, that node may temporarily choose another root until the exact authenticated connection returns or the origin's next signed renewal arrives (normally about 30–35 seconds). Relayed peers are not suppressed merely because their relay departs.

Open profile events are advisory setup diagnostics, not readable-overlay, replication-complete, or durable-write barriers. The callback runs inline and should enqueue/aggregate cheaply; exceptions from the new events are isolated from open correctness.

The deterministic regressions prove the stale-root, tracker-cohort recovery, shared-deadline, cancellation, bounded proactive-scan, join-drain, and ninth-provider cases. The exact downstream 500-file sequential cold-join benchmark still needs to be rerun after release to establish the new latency distribution.

## Validation

- full `@peerbit/pubsub` Node suite at final head: 243 passing, 0 failing
- full `@peerbit/shared-log` Node suite: 2,656 passing, 10 intentional pending
- focused root-query race suite: 73 passing
- focused provider-resolution suite: 8 passing
- focused fanout-open aggregate telemetry regression
- exact CI-small fanout gate: 20/20 joined, 400/400 messages delivered
- exact CI-loss fanout gate: 30/30 joined, loss target exceeded, no orphan state
- full workspace build
- repository lint: 0 errors; existing 15-warning ratchet unchanged
- SharedLog fence ratchet: 35 fields across 18 files
- diff and changeset checks

Patch changesets cover `@peerbit/pubsub` and `@peerbit/shared-log`.


peerbit-org:
Final-head follow-up after the earlier Fanout Gate failure:

- Root cause: `JOIN_REJECT_NOT_ATTACHED` from a responsive bootstrap tracker was treated as tracker failure. After all three trackers entered the exclusion cohort, an authenticated redial could return an excluded peer and the join stopped making progress.
- Final fix retains responsive trackers, rechecks exclusion after authenticated redial, rotates unhelpful cohorts after an unsuccessful round, and keeps proactive shard scans local, eight-way bounded, and under one shared 12-second discovery deadline.
- Exact local CI-small: 20/20 joins and 400/400 deliveries; CI-loss: 30/30 joins, the loss target exceeded, and no orphan state.
- Exact final-head pubsub suite: 243 passing, 0 failing. Full workspace build, lint (0 errors; the existing 15-warning ratchet), Changeset Guard, all split Node suites, both dependency-contract jobs, and Native/Rust all passed.

The P1-only diff is 3,266 additions / 266 deletions across 10 files. About two-thirds of additions are deterministic regression tests; production TypeScript is +1,129 / -229 (net +900). There are no new dependencies, lockfile changes, wire formats, persisted schemas, or migrations.

PR #1419, opened 2026-09-03
chore: version packages

peerbit-org:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and the packages will be published to npm automatically. If you're not ready to do a release yet, that's fine, whenever you add more changesets to master, this PR will be updated.

# Releases
### Patch Changes

- Updated dependencies []:
  - @peerbit/document@15.0.30
  - @peerbit/trusted-network@6.0.128
### Patch Changes

- Updated dependencies []:
  - @peerbit/document@15.0.30
### Patch Changes

- Updated dependencies []:
  - @peerbit/trusted-network@6.0.128
### Patch Changes

- Updated dependencies [[`991f5f8`](https://github.com/dao-xyz/peerbit/commit/991f5f8b86a16731e7063b3bd708e6811474e4c8)]:
  - @peerbit/document-interface@3.2.76
### Patch Changes

- [#1418](https://github.com/dao-xyz/peerbit/pull/1418) [`991f5f8`](https://github.com/dao-xyz/peerbit/commit/991f5f8b86a16731e7063b3bd708e6811474e4c8) Thanks [@peerbit-org](https://github.com/peerbit-org)! - Publish the document interface under a fresh npm version after the previous version was reserved but remained unavailable from the registry.
### Patch Changes

- Updated dependencies [[`991f5f8`](https://github.com/dao-xyz/peerbit/commit/991f5f8b86a16731e7063b3bd708e6811474e4c8)]:
  - @peerbit/document-interface@3.2.76
  - @peerbit/document@15.0.30
### Patch Changes

- Updated dependencies []:
  - @peerbit/document@15.0.30

PR #1420, opened 2026-09-03
test: harden receipt and dependency liveness checks

peerbit-org:
## Summary

- split the persisted-receipt legacy-confirm and stale-session checks so each case must inject its invalid evidence and observe a retry before cancellation
- pass explicit no-audit/no-fund config to the nested npm process used by the patch-package security fixture
- provision exact Node 18.20.8 with setup-node in CI and stable/RC release jobs, then pass its absolute executable to the crypto compatibility smoke
- remove implicit npx runtime downloads; the smoke now validates the executable before package work and fails closed when it is absent or wrong
- keep all product delivery deadlines, retries, wire behavior, and runtime behavior unchanged

## Why

The changeset-only recovery PR #1418 exposed a wall-clock false negative in the combined receipt test. The generated version PR #1419 then exposed two independent release-gate liveness dependencies: patch-package nested npm audit traffic and an implicit npx download of node@18. The latter consumed the full five-minute child deadline before the compatibility fixture ran.

The tests should prove their intended contracts without depending on unrelated audit-service latency, an overloaded 350 ms RPC window, or a hidden npm runtime bootstrap. Node 18 compatibility remains a real, fail-closed execution proof.

## Verification

- pnpm run test:security-dependencies
- pnpm run test:release-security-contracts
- packed isolated crypto consumer under exact Node 18.20.8
- missing/wrong Node 18 runtime fails before package work
- targeted Aegir run for both receipt tests
- 10 repeated targeted processes (20/20 cases), including six concurrent-pressure runs
- independent final diff review
- Prettier and git diff --check

No changeset is included because only tests, CI/release tooling, and documentation change.


peerbit-org:
Follow-up CI evidence on `ad17fa1b4`: Node 24 completed the unchanged strict release-security gate, while the concurrent Node 22 lane timed out only in `npm audit --omit=dev --json`. The passing Node 24 lane then logged its own advisory POST failure before pnpm retried successfully. Commit `1a93e17bc` therefore serializes only the two identical security lanes (`max-parallel: 1`) and adds a contract assertion. It changes no audit command, timeout, package graph, report parsing, or zero-vulnerability requirement; expected CI cost is roughly three minutes.

peerbit-org:
Follow-up after the isolated attempt-2 evidence:

- Node 22 passed the complete security gate.
- Node 24 passed the version-sensitive packed-consumer/runtime/crypto proof, including its strict zero-finding consumer audit.
- The later, graph-independent workspace production audit then exhausted pnpm's own bounded advisory-endpoint retries and timed out for a second time.

Commit 6afe10e32 keeps the gate fail-closed while removing that redundant request. CI now runs the compatibility/package proof on both Node 22 and Node 24, serialized, and runs the exact frozen-workspace production + full audits once on Node 22. Stable and RC publication still invoke the aggregate gate and therefore still require every proof plus both root audits before publishing. No audit finding is ignored or downgraded; the exact two existing documented workspace-only CVE exceptions are unchanged.

Local validation: release security contracts, dependency contracts, Prettier, and diff checks pass. An independent review found no correctness/security blocker.

peerbit-org:
The refreshed run exposed the same independent liveness defect in both serialized lanes: `npm audit --omit=dev --json` was killed by the smoke's 300-second process timeout. npm's default fetch timeout is also 300 seconds, so a stalled bulk-advisory POST consumed the whole parent budget before Arborist could execute its separate quick-audit fallback.

Commit 9f6d51793 sets only that advisory request timeout to 60 seconds while retaining the existing 300-second total process ceiling. This is not timeout inflation and does not suppress or retry an advisory result. A response must still exit 0, parse as audit report v2, contain an empty vulnerability map, and report total=0. npm fetch-retry flags are deliberately absent because its fetch layer does not retry POST requests; contract tests now pin that detail and all zero-finding assertions.

Both earlier lanes completed the contract and publication-closure phases before the transport timeout; 17 other substantive jobs were green. Local Prettier, release-contract, dependency-contract, and diff checks pass, and the change received an independent source-level review.

Issue #1417, opened 2026-09-03
TrustedNetwork v2: durable exact policy/fence readiness


peerbit-org:
## Corrected scope

This issue tracks a TrustedNetwork-v2-specific readiness gate. It does not add a generic SharedLog accepted-frontier API.

Review proved that generic log membership cannot establish durable application acceptance:

- Log.join can persist entries without signature verification;
- reopen and rollback restore indexed entries without a new admission decision;
- canAppend is stateful and cannot be replayed later as proof of the historical verdict;
- append, prepared/native, commit-only, and direct index/graph paths bypass _joining;
- program onChange completion cannot be reconstructed after a crash without program-owned durable state.

A 391-line prototype was held uncommitted and unpublished after executable tests found false acceptance of wrong local parent bytes, ordinary-owner outcome loss, non-prompt cancellation, and concurrent append callback races. It also added about 40.8 KiB unpacked across log and shared-log while remaining incomplete.

## Required contract

Given an exact named TrustedNetwork v2 policy EntryV0 CID:

1. fetch its bytes under one bounded deadline and caller signal;
2. recompute the CID and require equality with the requested CID;
3. authenticate canonical EntryV0 bytes, the sole policy-authority signature, and the network-bound policy body;
4. resolve only bounded authenticated policy ancestry;
5. ingest and publish through the crash-safe policy checkpoint;
6. require the durable current policy head to equal the named digest and sequence, not merely report accepted or duplicate; and
7. report readiness or invoke a supplied callback only while that exact durable head is protected by the current-lifecycle accepted-policy lease.

Where a resource fence is named:

1. independently recompute and authenticate the exact fence EntryV0 CID;
2. ingest it while holding the accepted-policy lease;
3. require the durable current resource-fence head to equal the named fence CID;
4. hold a stable resource-fence lease through the callback using policy-then-resource lock ordering; and
5. keep protected-log replay and operation revalidation as explicit additional gates.

An explicit empty resource frontier is valid only when encoded in the authenticated signed fence. An empty caller array is never proof that a remote is empty or current.

## Existing foundation

- v2 authority EntryV0 canonicalization and sole-authority signature validation;
- authenticated policy digest derivation;
- crash-safe policy ingest and checkpoint publication;
- bounded accepted-policy leases with lifecycle, timeout, and cancellation cleanup;
- resource-fence EntryV0 CID authentication;
- resource-fence reducer and crash-safe anchor held internally.

[...]

PR #1422, opened 2026-09-04
feat(trusted-network): add bounded exact policy-head lease

peerbit-org:
## Summary

- add an internal, bounded exact-CID resolver for trusted-network v2 policy entries
- add `withExactPolicyHead` to authenticate and serialize work against an already-ingested durable policy head
- distinguish exact wrapper-byte authentication from durable policy identity `(sequence, signed body digest)`
- document the partial protocol boundary and fail-closed limits

## Safety and bounds

- raw SHA-256 CIDs only, at most 128 characters
- policy entries at most 128 KiB
- ancestry and queued-operation bounds of 64
- hard 10-second acquisition ceiling; callers may only shorten it
- absent resolvers, unavailable bytes, invalid bytes, cancellation, and deadlines fail closed
- one background authentication slot prevents unbounded timed-out work

## Scope

This is a prerequisite for #1417, not completion of it. It only leases a policy head that has already been ingested and durably checkpointed. It does **not** fetch and publish a new durable head, activate v2 publicly, add resource fencing, or provide complete cold-peer revocation. Keep #1417 open for those follow-up slices.

## Verification

- full trusted-network package: 146 passing
- focused exact-policy/CID suite: passing
- package TypeScript build: passing
- `git diff --check`: passing
- independent correctness/security review: no blocking findings
- changeset included for `@peerbit/trusted-network`


peerbit-org:
Follow-up #1424 is now stacked on this exact head. It promotes the authenticated exact candidate through named admission and crash-safe checkpoint before callback; #1422 remains the small authentication/lease prerequisite.

PR #1421, opened 2026-09-04
fix(shared-log): recheck replicator membership until settle

peerbit-org:
## Summary

- continue checking the authoritative replication index on the existing recovery cadence after bounded V2/subscriber recovery attempts are exhausted
- preserve the existing network-attempt cap, outer timeout, role-age requirement, and fail-closed index predicate
- reject pending `waitForReplicator` calls promptly on log close and remove their timers/listeners

## Why

Replication transition events are wake hints, not replayable state. A transiently empty index read on the final transition wake could previously leave an otherwise-ready waiter parked until timeout. Current diagnostics captured an open/current session, active Full V2 receive state, mature unique membership, and a matching `replicator:mature` event while one waiter still expired.

This is intentionally separate from persisted-receipt readiness and from the downstream missing-conflict-head report.

Closes #1349.

## Verification

- deterministic regression fails on unpatched `origin/master` and passes here
- focused `waitForReplicator` suite: 37 passing
- exact close regression proves prompt `ClosedError` and no later index/network ticks
- package TypeScript build, ESLint, Prettier, changeset status, and `git diff --check` pass
- independent review found no remaining findings

## Compatibility

No public API, storage, or wire-format change. Runtime delta is small and uses the existing recovery timer; only local authoritative reads continue after the network-attempt budget.

PR #1423, opened 2026-09-04
fix(shared-log): repair unchanged successor sessions

peerbit-org:
## What changed

- Detect the first successfully committed, non-empty Full replication announcement from an authenticated same-identity successor session.
- When that Full reaffirms unchanged durable ranges, schedule the existing receipt-tracked join-authoritative repair and bypass only stale predecessor delivery hints.
- Bound recovery to once per exact PeerSession with lifecycle, admission, and receive-epoch fencing.

## Why

A same-key replacement can subscribe without a preceding unsubscribe. Its principal-owned ranges may already be present, so an unchanged Full previously produced no replication-change transition and no authoritative catch-up. Stale predecessor delivery hints could then suppress a missing sibling head indefinitely. This is distinct from #1421: that PR makes `waitForReplicator` resilient to missed membership wakeups; this PR restores missing-entry convergence after an unchanged successor Full.

Downstream evidence is consistent with this gap, but did not distinguish raw SharedLog absence from Documents projection absence, so this PR does not overclaim the exact downstream root cause.

## Validation

- New regression failed deterministically before the runtime change: the middle sibling hash never reached the synchronizer.
- The focused regression passes after the fix and covers replay, empty Full, stale session, no synthetic public events, and targeted hint bypass.
- Complete SharedLog events suite: 89 passing.
- SharedLog build, Prettier, ESLint, and `git diff --check`: pass.
- Independent concurrency/lifecycle review: no blocker.

## Compatibility and size

No public API, wire-format, storage-format, or dependency change. Runtime growth is a WeakSet plus the bounded reuse of existing repair machinery; no polling loop or per-write path is added. A patch changeset for `@peerbit/shared-log` is included.

## Release sequencing

Keep this behind the currently staged #1419 cohort release. Then land this with #1421 before cutting the next coherent downstream test cohort.


peerbit-org:
Additional downstream corroboration (published 5.3.38 cohort, so neither this branch nor #1421 is present):

- peerbit-examples run 33797823722: the two built-in attempts failed at different points—one waitForReplicator timeout, and one restarted three-writer session converged with only 2 of 3 expected conflict heads.
- peerbit-examples run 33797859563: one attempt never exposed persistedReceiptPeerSession(remoteHash) and another timed out in waitForReplicator.

Raw runs:
- https://github.com/dao-xyz/peerbit-examples/actions/runs/33797823722
- https://github.com/dao-xyz/peerbit-examples/actions/runs/33797859563

The waiter timeout is addressed separately by #1421. The missing sibling head is the failure shape modeled by this PR’s unchanged-successor regression. The absent receipt-session observation may be an earlier capability/session symptom and is not claimed fixed by this PR alone. Acceptance remains an unchanged downstream rebaseline on the coherent release containing both #1421 and this PR.

peerbit-org:
Combined-stack verification update: a clean integration simulation of #1419 → #1421 → #1423 completed the full `@peerbit/shared-log` suite with 2,677 passing / 10 pending. This includes the new waiter-recheck and same-identity successor-repair regressions alongside persisted receipts, V2 state, replication, and pruning coverage.

The planned release remains one coherent downstream cohort containing both #1421 and #1423; neither fix should be validated through a mixed dependency graph.

peerbit-org:
Independent review + stress update:

- no correctness, fanout, session-fencing, compatibility, or bloat blocker found;
- the repair is admitted only after an authenticated exact-current V2 Full commits durably, and is once per exact successor PeerSession;
- it reuses the existing per-target coalesced, receipt-tracked repair frontier and adds no per-write work;
- runtime growth is 43 lines / 1,788 source bytes with no public API, wire, storage, export, lockfile, or dependency change;
- the #1421 + #1423 focused regressions passed together in 20/20 fresh process runs, in addition to the combined 2,677-test package pass.

The upstream regression proves scheduling of the stale-hint-suppressed sibling; unchanged downstream multi-process/cross-platform testing remains the end-to-end convergence proof.

PR #1427, opened 2026-09-05
fix(shared-log): preserve receipt confirmation across delayed commits and sessions

marcus-pousette:
Persisted delivery can stop making progress on a quiet log even after readiness preflight succeeds. A confirmation query can arrive before the receiver commits its replication state; the old retry path replaced the outstanding sequence and sent another Full before each query, repeatedly closing the receiver's admission gate. Session changes with a missed wake and loss of outbound sender state after preflight exposed two further recovery gaps.

This change retains one exact outstanding sequence/revision, uses query-only retries between progressively spaced Full repairs, rechecks session and replicator identity on the readiness recovery tick, and rearms missing sender state inside persisted delivery itself. Full repair remains available for updates acknowledged by transport but missed by host admission. Temporary receive reservations preserve the existing confirmation waiter. Remote Full rearm also backs off within its existing bounded generation, leaving slow commits and delayed Applied responses time to settle before another challenge rotation.

Authenticated peer/session/challenge binding, fresh entry leader plans, committed-hash receipts, cancellation and the caller's deadline remain mandatory. No timeout increase, wire change, public API change, dependency addition or downstream-specific code is introduced.

The new Windows gate also caught a real native persistence defect: named coordinate flush on offline reopen used a read-only file handle, but Windows requires write access for fsync. Existing regular files and atomic-replacement temporary files now open with non-truncating `r+` before sync. Missing files are not created; real file-sync errors remain fatal; directory behavior is unchanged. This has its own native-backbone patch changeset.

Validation:

- Deterministic failures captured against the previous sender: delayed outstanding ACK, actual receiver reservations longer than the retry interval, overlapping Full/resync recovery, and older/newer revision fairness.
- Deterministic missing-event session replacement and missing sender state after successful preflight also failed before their fixes.
- 251 receipt, sender/receiver and replicator tests pass locally with zero retries, including disk-backed quiet recovery and propagation of three independent heads.
- A real paired sender/receiver/rearm-coordinator regression with a ten-second apply delay, one-second confirmation cadence and delayed response failed under the fixed rearm cooldown and now completes within the unchanged sixty-second deadline. Cooldown/reset and abort-ignoring per-session/global bounds are covered.
- Added the same receipt/session CI gate on Ubuntu, macOS and Windows with retries disabled.
- The new Windows job exposed a preexisting Unix-only root postinstall command before tests could run; replaced that test-only Chai setup with the equivalent portable Node script. The initial draft CI is superseded by this substantive follow-up, not rerun blindly.
- The initial macOS gate exposed two transport-delivery fixtures that mistook local full coverage plus pubsub subscription for remote replication ownership. A gated-apply regression reproduces that gap; both fixtures now use the existing exact remote-role fence without changing their transport assertions or timeout. All 15 delivery-file tests pass with zero retries; no runtime bypass was added.
- The subsequent Windows gate passed 250/251 tests and exposed the read-only fsync defect above. Four deterministic actual-file tests captured the failing mode and now verify writable flush, preservation, absent-file rejection, exact EPERM/EIO propagation, handle cleanup and no atomic publication after failed sync. The native index plus focused suite passes 115/115 with zero retries and independent review; fresh real Windows CI remains required.
- Packed comparison against shared-log 16.0.28: +2,487 compressed bytes / +11,116 unpacked bytes; unchanged 208 files.
- Native-backbone matched local base/candidate pack comparison: +175 compressed / +572 unpacked bytes; unchanged 66 files. No Rust or dependency changes.

Downstream evidence: [isolated builder failure](https://github.com/dao-xyz/peerbit-examples/pull/329#issuecomment-5549244470), [cohort rebaseline](https://github.com/dao-xyz/peerbit-examples/pull/300#issuecomment-5549251564). Their fresh unchanged three-OS lifecycle run remains the downstream acceptance gate after release.


marcus-pousette:
The receipt/session gate is now green on **all three operating systems on the same exact head `07effe83bc1e3fa63c0b29536032d131b154efe2`**, first attempt, **251 tests per OS with `--retries 0`**:

- [Windows](https://github.com/dao-xyz/peerbit/actions/runs/33953271324/job/101272642726)
- [Ubuntu](https://github.com/dao-xyz/peerbit/actions/runs/33953271324/job/101272642734)
- [macOS](https://github.com/dao-xyz/peerbit/actions/runs/33953271324/job/101272642746)

Windows explicitly passes the native receipt/offline-reopen case that previously failed on read-only fsync access. The same-sequence precommit and slow-apply/watchdog regressions also pass. No protocol timeout inflation, skipped receipt test, or workflow rerun was used to obtain this result; earlier failed revisions and their substantive fixes remain documented above.

The complete CI matrix is still a merge gate. This is upstream evidence, not a claim that downstream's unchanged lifecycle acceptance suite has run on an unpublished cohort. The release PR remains held until this source revision is fully validated and merged.

marcus-pousette:
Published and verified the receipt/session liveness patch cohort from release commit `80901d52cb8081a5f8cac592a5921f06d68578e8` via #1430.

Core pins:

| Package | Published version |
| --- | --- |
| `peerbit` | `5.4.1` |
| `@peerbit/document` | `15.0.32` |
| `@peerbit/shared-log` | `16.0.29` |
| `@peerbit/native-backbone` | `0.2.14` |
| `@peerbit/trusted-network` | `6.0.130` |

Relevant coherent sibling pins: `@peerbit/program@6.0.59`, `@peerbit/pubsub@5.4.5`, `@peerbit/blocks@4.3.0`, `@peerbit/blocks-interface@2.2.0`, `@peerbit/stream-interface@6.0.16`, `@peerbit/indexer-sqlite3@3.0.20`, `@peerbit/crypto@3.1.6`, `@peerbit/log@6.2.32`, and `@peerbit/rpc@6.1.27`. Test consumers can use the newly published `@peerbit/test-utils@3.1.40`.

The repair preserves outstanding confirmation sequences across delayed commits, follows replacement sessions and lost replicator state, and recovers missing outbound sender state during actual persisted delivery. Full repairs remain bounded and spaced so slow commits can settle. Windows native coordinate fsync now uses non-truncating writable regular-file handles. No wire/public API change, dependency addition, caller-timeout increase, or relaxed receipt/authorization condition is included.

Verification:

- Full source and generated version CI passed. The source post-merge CI passed as well.
- On both source and version candidates, the receipt regression gate passed **251 tests on each of Ubuntu, macOS, and Windows**, first attempt, with retries disabled. The suites have no local retry override. The original disk-backed native receipt/offline-reopen case passed on each platform.
- [Generated version CI](https://github.com/dao-xyz/peerbit/actions/runs/33954778004), [successful publication](https://github.com/dao-xyz/peerbit/actions/runs/33956341867), and [successful post-release automation](https://github.com/dao-xyz/peerbit/actions/runs/33957033974).
- All **16 exact published versions** were checked against the release plan, with integrity metadata, expected internal dependency anchors and no unresolved workspace protocols. All 16 remote tags dereference to the release commit above.
- A fresh private consumer using one `pnpm install --ignore-scripts` and 14 exact cohort pins passed a filesystem-only resolution check: **14 package instances and 55 declared pinned dependency edges**, with shared Program/pubsub/blocks import paths across consumers. This is module-resolution evidence, not execution of downstream replication, fsync or disposal tests. No `dedupe` or security audit was run for this check.

Downstream acceptance remains a fresh unchanged, zero-retry Ubuntu/macOS/Windows lifecycle run on this cohort. Pin the complete set of Peerbit packages the workspace uses, perform one plain install, refresh #300, and retain raw first-attempt evidence before merging held runtime/release work. Upstream regressions do not substitute for that downstream acceptance run. Readiness remains preflight; actual persisted receipts remain the remote durability proof.

Separate scope: #1429 implements opt-in required batching and explicit local commit outcomes, but is **not published in this patch cohort**. Its integrated Documents suite passed 541 tests locally with retries explicitly forced to zero, plus 89 focused batching/durability tests, and fresh CI is running on the versioned base. It does not promise an all-or-none transaction. The trusted-network resource-fence groundwork remains internal and does not constitute complete public cold-peer revocation.

PR #1429, opened 2026-09-05
feat(document): require batching with explicit local commit outcomes

marcus-pousette:
Add opt-in `Documents.putMany(documents, { batching: "required" })` for callers that need a real supported independent batch and must not silently receive sequential puts.

Required mode captures invocation order, encoded document values, all byte-view keys and supported append options before asynchronous work. Duplicate keys and unsupported document modes/options fail before handoff. Existing default behavior and successful result shape remain unchanged; ordinary calls gain no serialization/decode work.

`DocumentBatchCommitError` separates local append state from projection and remote receipt failure:

- `not-started`: local append was not handed off; safe to replay that append, not arbitrary application/policy side effects.
- `committed`: complete trusted append evidence maps to immutable `{ index, hash }` pairs in captured input order. Do not replay merely because later projection or remote delivery failed.
- `indeterminate`: lower preparation/append may have mutated state without a complete acknowledgment, or native durability failed. Recovery is required; an empty evidence array never proves an empty commit.

This is **not** an all-or-none transaction, an exact fsync-count guarantee, or a substitute for persisted delivery receipts. Partial/contradictory evidence cannot invent input indexes. A remote delivery failure preserves the original `PersistedDeliveryError` as the cause.

Validation:

- Required-batching suite: 20 passing, including input/options/key-view mutation, no fallback, unsupported and duplicate inputs, cancellation, exact post-commit evidence and ambiguous lower outcomes.
- Real native durable-barrier regression: 2 passing (legacy and required modes), retaining one batch mirror barrier and unsafe-replay classification on failure.
- Existing persisted-delivery suite: 29 passing.
- Independent review caught and fixed unsafe classification of a lower `undefined` return and caller-owned non-Uint8Array key aliases; both have regressions.
- TypeScript, ESLint, formatting and diff checks pass.
- Package delta: +5,134 compressed / +18,154 unpacked bytes, including +4,937 runtime JS bytes; no dependencies.

The additive public API has a minor changeset. This is isolated from the P0 receipt/session fix in #1427 and must not delay that fix if broader validation identifies further work here.


marcus-pousette:
Additional validation on exact head `783c87d3fd03041a0301dc29316329e8eb6aef2c`: the complete Documents Node suite passed, **541 tests with zero retries**, in approximately nine minutes. This includes durable/SIGKILL recovery, existing replication and authorization behavior, index/search behavior, and the new required-batching regressions.

Command: `node scripts/ci/assert-tests-ran.mjs node ./node_modules/aegir/src/index.js run test --roots ./packages/programs/data/document/document -- -t node --no-build -- --retries 0` after the Documents package build. Process exited 0; the non-empty-suite assertion reported 541 passing. CI remains a separate merge gate.

marcus-pousette:
Rebased onto the merged receipt fix and generated P0 release metadata at `80901d52cb8081a5f8cac592a5921f06d68578e8`. Current PR head is `15e511ab6bda3b4377708df1f6feb5c3aa20ffc7`; the single implementation patch is unchanged (range-diff equality), with exactly the original seven files and only this PR's required-batching changeset.

Local integration validation on the merged P0 source:

- Required-batching and native durable-commit suites: **89 passing** with zero retries.
- Complete Documents suite: **541 passing**, nine minutes, with retries explicitly disabled for every discovered test.
- Native-backbone, SharedLog, and Documents outputs were rebuilt before testing.

Evidence correction: the older full-suite invocation used only CLI `--retries 0`, but the existing domain suite sets its own retry count. This new local run used an external Mocha root hook to force all 541 tests to zero retries after suite construction and reject any nonzero current retry. No tracked tests, timeouts, or production code were changed for that verification. All tests passed without a retry. The subsequent rebase changed only generated release metadata, not runtime or test source.

Fresh CI is running on this exact versioned-base head. This remains a separate additive minor API change and is **not** included in the P0 patch cohort. It does not claim an all-or-none transaction or equate local append evidence with projection completion or remote durability.

peerbit-org:
Superseded by #1431 under the repository automation identity. The required-batching patch is unchanged; #1431 has a fresh bot-authored commit on corrected master and will require its own green CI before merge. This PR is preserved as historical context.

PR #1446, opened 2026-09-07
perf(log): bound unrelated CUT coverage planning work

### LINKED ISSUE #1286: Scale shared-log to 100M global entries with bounded per-peer state

peerbit-org:
## Summary

Small measured slice toward #1286: use existing reverse adjacency for CUT coverage checks when cheaper, retaining the previous amortized head scan for wide relevant batches. No new retained index, wire/storage/API change, or tombstone deletion.

Coverage predicates and missing-parent/reset semantics remain unchanged. Five differential tests cover all three planners, exhaustive predicate combinations, replacement/deletion/head transitions and rebuild.

## Measurements and validation

Seven alternating within-process Rust microbenchmark samples: with 100k unrelated same-gid CUT heads, a fixed 128-entry batch changed from median 27.84039ms to 0.02273ms. This is graph planning only, NOT full cold-join/reopen CPU or a compaction claim. The 100k current-head APPEND control was about 4.6% slower in this run. An initial unconditional reverse lookup badly regressed that control; its failed evidence was retained and motivated the count-based cost gate.

- Rust: 43 log tests and 58 backbone tests passed; both WASM target checks and formatting passed.
- Rebuilt JS/WASM wrappers: 42 log-rust and 111 native-backbone tests passed with zero retries/pending, strict unhandled rejections, natural exit.
- Both pnpm-packed/extracted artifacts: 63 assertions, exact binary and isolated-resolution proof, including embedded raw-receive CUT retention.
- Existing backbone checkpoint fixture emitted a FileHandle GC warning; preserved, not attributed to this patch.

Full methods, limitations and raw artifact hashes: `packages/log/rust/benchmark/CUT_COVERAGE.md`.

Both `@peerbit/log-rust` and `@peerbit/native-backbone` receive direct patch changesets because both embed this graph. This does not close #1286 or authorize logical history truncation. Cross-platform CI and release remain pending.


peerbit-org:
Superseded by the exact reviewed patch integrated in #1451, now merged as bf3b3164d675b736ee8f6f4965bcafc71ad22751 after final combined-tree CI 34118306820 passed on its first attempt. Original branch, commits, and CI evidence are preserved. Closing this PR avoids duplicate integration; publication is still pending the generated version PR #1450.

PR #1463, opened 2026-09-09
fix: integrate pubsub lifecycle cleanup and bounded query diagnostics

peerbit-org:
## Purpose

Integrate the already reviewed upstream changes from #1460, #1461 and #1462 into one exact candidate for full CI and a coherent Changesets release. All original commits are preserved by normal merges; this PR introduces no additional source edits.

- Bound ephemeral pubsub idle timers to their exact channel, timer and lifecycle. Pending publish/join work cannot rearm them during stop or touch a replacement channel.
- Add opt-in, invocation-local Documents query and RPC timing through the existing diagnostics surface: bounded detail events, a reserved terminal, sampled peer metadata, and isolated observer failures.
- Replace two terminal-only `tty-table` callers with already-locked `cli-table3`/`wrap-ansi`, removing their unused runtime CSV parser chain. Exclude compiled server tests from its published package.

## Compatibility and scope

26 paths, three original changesets, six directly affected packages. RPC/diagnostics receive additive minor changes; document/pubsub/server/test-utils are patches, with the normal dependent release cascade left to Changesets.

No changed wire/storage format, authorization, query coverage, timeout, receipt accounting, local commit contract or disposal proof. Diagnostics are off without a callback and are not a latency fix or durability proof. Terminal output retains content/ANSI/Unicode wrapping; byte-identical column layout is not promised. The dependency cleanup is not a claim that default Node consumer or React Native installation bloat is solved.

The controlled idle-timer regression does not identify every historical native/RPC shutdown tail. The immutable precommit lookup experiment remains separate from production mutable Shared FS stores and persisted receipt settlement. No downstream repository is modified.

## Validation

- Exact combined committed-tree guard passes against master `cef80ff755be13993e7076fc62a2ee87a29789a5`, with precisely the six affected packages covered.
- Original #1460 full first CI passed, including its 27 diagnostic cases. Original #1461 completed three-OS logs contain all nine new timer controls; original #1462 completed logs contain all six renderer cases and its actual advisory scan reports zero production findings. These individual results are not substitutes for combined-tree CI.
- Candidate-local rebuilt TypeScript and the 151 focused/existing local controls are being checked with strict unhandled rejections, retries disabled and natural process exit; their final outcome will be recorded before merge.
- A fresh full CI/guard/advisory run is required for this exact combined head. No workflow rerun, weakened gate, timeout inflation, manual version bump or direct publication is requested.

After this candidate is green and merged, the normal bot-generated version PR will be reviewed and released. No unpublished npm versions are asserted here.

PR #1465, opened 2026-09-09
fix(pubsub): settle route lookup resources on teardown

peerbit-org:
## Summary

Settle route lookups when their exact fanout channel stops, closes, detaches or is kicked. Release owned timers and abort listeners on every local settlement path; prevent stale local timeout/send callbacks from completing replacement proxy searches. Refuse new proxy work after channel/service shutdown. Preserve coalesced callers' independent deadlines and existing route-cache/reply policy.

This is a small resource-ownership fix, not a complete shutdown or receipt-liveness fix. No dependency, wire format, authorization, receipt or timeout-setting changes. Includes a pubsub patch changeset and 17 deterministic lifecycle cases in the existing zero-retry three-OS ownership lane.

## Validation and draft blocker

- TypeScript build and all 17 new component cases pass with strict unhandled rejections, zero retries and natural process exit.
- The identical final tests against the unchanged base module produce 4 passing / 13 failing cases. Thirteen assertions are not thirteen independent defects. These controlled-send ownership fixtures assert cleanup before their own bounded teardown; they are not end-to-end convergence or shutdown proof.
- A broader selection of 18 existing controls produced 17 passing / 1 failing: `re-parents when its parent disconnects`, after the relay stops. One explicitly matched baseline-only diagnostic reproduces the same assertion without this patch. The failure remains unresolved; this draft must not be merged on the strength of a retry or the focused tests alone.
- A separate unchanged RPC39 observation still leaves two referenced WebSocket connections after awaited test cleanup and exits naturally around 80 seconds. Its final observer records were truncated and diagnostic completeness failed; useful complete per-test records are retained. This patch is not claimed to fix that longer tail or the historical downstream native-cleanup hang.

## Deliberate boundaries

Unsolicited/late wire-reply cache learning and expected-responder policy are unchanged. Exact local callback ownership is not a protocol-epoch solution for reused wire IDs. No transport is disabled, no deadline is inflated, no forced-success exit or global socket cleanup workaround is added.

Keep this draft separate from merged #1463 and its generated release PR #1464 until the reconnection control is explained and the final candidate passes the appropriate checks. Downstream durability/disposal gates are not cleared by this PR.


peerbit-org:
Follow-up evidence, kept separate from this draft's passing route-ownership tests:

Two deterministic real-loopback component cases against the unchanged locked `@libp2p/websockets@10.1.19`, `@libp2p/utils@7.3.2`, and Node24.13.1 now reproduce cleanup defects:

1. Gate the server upgrade response, abort the real transport dial, wait for its `ConnectionFailedError`, then release the handshake. The same native WebSocket becomes OPEN after the rejected dial; the upgrader is never reached. `_connect`'s aborted catch branch throws before its close attempt.
2. Wrap an established native WebSocket with the actual `webSocketToMaConn`, then call `abort`. Abstract state becomes aborted, but the native socket remains OPEN; logging records `InvalidAccessError: invalid code` from `close(1006)`.

Both cleanup expectations failed in one bounded invocation (natural exit1,312ms; no retry/watchdog kill). Ordinary no-code close then reached CLOSED as an explicit fixture cleanup control, followed by recorded owned-server cleanup. This is mechanism-level evidence, not proof of which path caused the full RPC39 tail or the re-parent failure.

An integrity-verified comparison of the published WebSockets10.1.21 tarball found byte-identical relevant source/built paths. Its newer Utils7.4.1 adds end-event handling but no direct physical-close fallback after reset throws; indirect full-stack effects are untested. No blind dependency upgrade, transport disablement, global socket workaround or external repository change was made.

This PR remains draft and its broader reconnection control remains unresolved. #1463's independent integration checks passed before merge; release#1464 does not include this draft and must not be described as fixing these newly isolated transport defects.

